PR #2, opened 2019-04-12
Apply pending changes to  v1.0 articles

fujimotos:
This is a collection of patches for v1.0 articles since https://github.com/fluent/fluentd-docs/commit/6dad0eab466889ccae9972145f7e9054680a1c15

Since the directory layout has been changed, I have applied
each patch manually one by one, fixing their contents to fit the
GitBook style formatting.

With this, the v1.0 documentation should sync with v1.4.2.


edsiper:
thanks

PR #50, opened 2019-06-18
Add tcp and http doc

ganmacs:
docs for https://github.com/fluent/fluentd/pull/2395 and in_tcp

@repeatedly could you review this, please?
cc. @edsiper

### DIFF
--- a/plugins/input/http.md
+++ b/plugins/input/http.md
@@ -164,6 +164,27 @@ the following example).
 Respond with an empty gif image of 1x1 pixel (rather than an emtpy
 string).
 
+### &lt;transport&gt; section
+
+| type | default | available values | version |
+|:-----|:--------|:-----------------|:--------|
+| enum | tcp     | tls      | 1.5.0  |
+
+This section is for using SSL transport.
+
+```
+<transport tls>
+  cert_path /path/to/fluentd.crt
+  # other parameters
+</transport>
+```
+
+See "How to Enable TLS Encryption" section for how to use and see
+["Configuration example" in "Server Plugin Helper" article](/developer/api-plugin-helper-server.md#configuration-example) for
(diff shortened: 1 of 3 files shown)


plugins/input/http.md
```diff
@@ -327,6 +348,45 @@ curl -X POST -F 'json={"message":"foo+bar"}' http://localhost:9880/app.log
...
+Since v1.5.0, `in_http` support TLS tranport. Here is configuration example with HTTPS client.
+
+- in_tcp
+
+```
+<source>
+  @type http
+  bind 0.0.0.0
+  <transport tls>
+    insecure true
```
repeatedly: insecure example is not good.

plugins/input/tcp.md
```diff
@@ -170,6 +187,26 @@ TCPSocket.open('127.0.0.1', 5170) do |s|
   s.write('{"k":"v2"}' + "\n")
 end
 ```
+## Tips
+
+### How to Enable TLS Encryption
+
+Since v1.5.0, `in_tcp` support TLS tranport. Here is configuration example.
```
repeatedly: in_tcp seems to support TLS from further back so removing `since v1.5.0` is enough.

ganmacs: https://github.com/fluent/fluentd-docs-gitbook/pull/50/commits/d3fd3fa5f1fb91d9f942f55dd11fb13ece3a34dc fixed

plugins/input/http.md
ganmacs: fixed not to use insecure directive.
https://github.com/fluent/fluentd-docs-gitbook/pull/50/commits/96f993207361a4297bc103c9094aa1507c29f6e6

PR #53, opened 2019-06-28
update image markdown link to make it show in website

athom:
the image is not show up in webpage https://docs.fluentd.org/buffer
This PR try to fix it

image

### DIFF
--- a/plugins/buffer/README.md
+++ b/plugins/buffer/README.md
@@ -39,11 +39,7 @@ where chunks wait before the transportation. Every newly-created chunk
 starts from *stage*, then proceeds to *queue* in time (and subsequently
 gets transferred to the destination).
 
-<div>
-
 [![](/images/fluentd-v0.14-plugin-api-overview.png)](/images/fluentd-v0.14-plugin-api-overview.png)
-
-</div>
 
 
 ## Control Retry Behaviour


ganmacs:
Thank you! looks good! 
Could you add your sign to commit? https://github.com/fluent/fluentd-docs-gitbook/pull/53/checks?check_run_id=157746282

plugins/buffer/README.md
```diff
@@ -41,7 +41,7 @@ gets transferred to the destination).
 
 <div>
 
-[![](/images/fluentd-v0.14-plugin-api-overview.png)](/images/fluentd-v0.14-plugin-api-overview.png)
+[![](/images/fluentd-v0.14-plugin-api-overview.png)](https://github.com/fluent/fluentd-docs-gitbook/blob/1.0/images/fluentd-v0.14-plugin-api-overview.png)
```
ganmacs: Is`` correct?

athom: actually I don't understand why  this format `[x](y)`, 
can't this `x` work?

plugins/buffer/README.md
```diff
@@ -41,7 +41,7 @@ gets transferred to the destination).
 
 <div>
```
ganmacs: I found how to fix it.  Removing this `<div>` tag and L46 `</div>` tag will fix.

plugins/buffer/README.md
ganmacs: `[x](y)` means an image with a link like `<a href="/path1"></a>`.

plugins/buffer/README.md
athom: ah, thanks for pointing out.
I updated the PR.

ganmacs:
thank you!

PR #55, opened 2019-07-01
Fix the default value of queued_chunks_limit_size

summerwind:
This PR fixes the default value of `queued_chunks_limit_size` based on following PR.
https://github.com/fluent/fluentd/pull/2173


repeatedly:
Thx!

PR #64, opened 2019-07-09
Correct config for tls_client_cert_path

Schnitzel:
there is a small typo: `tls_client_cert_pat` is `tls_client_cert_path`


repeatedly:
Ah, good catch. Thanks!

PR #69, opened 2019-07-11
Added code block around embedding Ruby example

cpedro:
- There were no code block backticks around example code, which was causing it to not display correctly in the docs.


ganmacs:
Thank you!

PR #71, opened 2019-07-14
Make :udp a symbol

candlerb:
`default: :udp` instead of `default: udp` - which would fail with `NameError (undefined local variable or method 'udp' for main:Object)`


repeatedly:
good catch!

PR #74, opened 2019-07-16
Added secondary output example for buffers

cpedro:
Signed-off-by: Chris Pedro <chris@thepedros.com>


repeatedly:
Ah, we forgot to add example. Thanks!

PR #90, opened 2019-07-26
Remove Hadoop version 1 and CDH3/4 as outdated

romainr:
As of today, it is Hadoop 3 and CDH6.
Also fix Cloudera's Download page URL.


repeatedly:
Thanks :)

PR #92, opened 2019-07-29
Fix parameter name of default value

summerwind:
This PR fixes the wrong parameter name of queued_chunks_limit_size's default value.


repeatedly:
Sorry for the delay.

PR #101, opened 2019-09-04
Add private_key_path and private_key_passphrase parameters

cosmo0920:
I've found that mutual authentication parameters are not documented.

Signed-off-by: Hiroshi Hatake <hatake@clear-code.com>


repeatedly:
Looks good.

PR #111, opened 2019-10-25
Add notes logrotation on Windows

cosmo0920:
Windows does not permit delete and rename files which are owned by another process.
And Fluentd v1.7.4 will split log files when logrotate feature is
enabled on Windows.

Signed-off-by: Hiroshi Hatake <hatake@clear-code.com>


repeatedly:
Thanks for quick update.

cosmo0920:
Thanks!

PR #113, opened 2019-11-10
Fix some broken links

nokute78:
I tried to fix two broken links.

* `python logger` in https://docs.fluentd.org/how-to-guides/raspberrypi-cloud-data-logger
* `the grep plugin` in https://docs.fluentd.org/how-to-guides/parse-syslog


repeatedly:
Good catch. Thanks!

PR #120, opened 2019-12-11
fix some links 

nokute78:
I fixed some broken links.

* `v0.12 compatibility` of https://docs.fluentd.org/configuration/buffer-section
* `json, ltsv and regexp` of https://docs.fluentd.org/configuration/parse-section


ganmacs:
Thanks!

PR #124, opened 2019-12-12
Add sd

ganmacs:
ref https://github.com/fluent/fluentd/pull/2541

PR #127, opened 2019-12-13
Follow up add sd

ganmacs:


plugins/service_discovery/static.md
```diff
@@ -13,7 +13,9 @@ It defines the list of targets. it is similer to the server directive in out_for
 
   <service_discovery>
     @type static
-    host 127.0.0.1
+    <server>
```
ganmacs: I was wrong about it. I changed it to support multiple servers.

https://github.com/fluent/fluentd-docs-gitbook/pull/124#discussion_r357016913

PR #128, opened 2019-12-13
Add a new config option and helper methods

ashie:
* strict_config_value
  * use_nil
  * use_default

ref: https://github.com/fluent/fluentd/pull/2685


repeatedly:
Thanks!

PR #130, opened 2019-12-19
Fix broken slides link in plugin-development

yusong-shen:
Fix broken slides link in gitbook at https://docs.fluentd.org/plugin-development#overview


repeatedly:
Thanks!

PR #136, opened 2020-01-07
estimate_current_event field default value is not sync

JiHyunSong:
default value is true,
ref : https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/parser.rb\#L94


ganmacs:
thank you!

PR #139, opened 2020-01-13
fix: Change Aapche to Apache

joestrouth1:
On `/installation/post-installation-guide`, the word Apache was spelled incorrectly.


repeatedly:
Ah, good catch!

PR #140, opened 2020-01-15
Fix display of td-agent.conf file on docs site.

AlexHeylin:
Renders OK here as was, but fails on docs site and it renders the XML to plain text which is useless and very confusing.


AlexHeylin:
This seems to not have the fix in - PLEASE DELETE. I'll resubmit.

PR #141, opened 2020-01-15
Fix Syslog to Graylog example

AlexHeylin:
Fix formatting.
Correct & clarify content.

### DIFF
--- a/guides/graylog2.md
+++ b/guides/graylog2.md
@@ -98,12 +98,14 @@ Configure `/etc/td-agent/td-agent.conf` as follows.
 ```
     <source>
       type syslog
+      bind 0.0.0.0
+      port 5140
       tag graylog2
     </source>
 
     <match graylog2.**>
       type gelf
-      host 0.0.0.0
+      host 127.0.0.1
       port 12201
       <buffer>
         flush_interval 5s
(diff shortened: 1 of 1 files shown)


guides/graylog2.md
```diff
@@ -95,21 +95,24 @@ We also need to gem-install GELF's Ruby client.
     $ sudo /usr/sbin/td-agent-gem install gelf
 
 Configure `/etc/td-agent/td-agent.conf` as follows.
-
+```
     <source>
       type syslog
+      bind 0.0.0.0
```
ganmacs: I think we don't need to write a default value because this is an example and shorter is better.

ashie:
The current document is here: https://github.com/fluent/fluentd-docs-gitbook/blob/1.0/how-to-guides/graylog2.md

* Quotes are already added
* Default values shouldn't be added as ganmacs said:
  > I think we don't need to write a default value because this is an example and shorter is better.

The only thing we need to pickup from this PR is fixing destination host address:

```diff
-      host 0.0.0.0
+      host 127.0.0.1
```

`0.0.0.0` is invalid as a destination address.
We'll do it in another PR.
Anyway, thank you for your contribution.

ashie:
#313

PR #156, opened 2020-02-21
update gem install doc to reflect Ruby 2.4 support

clacsinaRX-M:
Signed-off-by: Christian Lacsina <christian.lacsina@rx-m.com>

Ruby Gem installation docs still list Ruby 2.1 as the minimum version. Minimum version is now Ruby 2.4 according to https://www.fluentd.org/blog/drop-schedule-announcement-in-2019


ganmacs:
Thanks!

PR #159, opened 2020-03-05
Fix geoip code example

montac:
I noticed that, the code example in `https://docs.fluentd.org/filter/geoip` about the use case of `country_code` starts with a `<filter>` tag but closes with `</match>` tag.


ganmacs:
Could you follow DCO https://github.com/fluent/fluentd-docs-gitbook/pull/159/checks?check_run_id=488108989 ?

repeatedly:
Sorry for the delay. Just merged.

PR #162, opened 2020-03-20
Fix document path for parser filter

wreulicke:
`parser` filter plugin is already included in fluentd core.
So, I fix document path for parser filter plugin.


ganmacs:
Thank you!

PR #166, opened 2020-03-31
Add example of 'remove_keys' including nested keys

kir4h:
This PR adds an additional example for `record_transformer`. It covers the usage of the `remove_keys`, including the removal of a nested key.

### DIFF
--- a/plugins/filter/record_transformer.md
+++ b/plugins/filter/record_transformer.md
@@ -59,6 +59,40 @@ into
 {"total":100, "count":10, "avg":"10"}
 ```
 
+This filter can be also used to remove fields through the `remove_keys`
+parameter (including nested keys, by means of `jsonpath` syntax)
+
+Given the filter:
+
+```
+<filter foo.bar>
+  @type record_transformer
+  remove_keys hostname,$.kubernetes.pod_id
+</filter>
+```
+
+And a message:
+
+```
+{
+  "hostname":"db001.internal.example.com",
+  "kubernetes":{
+    "pod_name":"mypod-8f6bb798b-xxddw",
+    "pod_id":"b1187408-729a-11ea-9a13-fa163e3bcef1"
+  }
+}
+```
+
+The output would be:
+
+```
+{
+  "kubernetes":{
+    "pod_name":"mypod-8f6bb798b-xxddw"
+  }
+}
+```
+
 With the `enable_ruby` option, an arbitrary Ruby expression can be used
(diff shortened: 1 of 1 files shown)


plugins/filter/record_transformer.md
```diff
@@ -59,6 +59,40 @@ into
...
+The output would be:
+
+```
+{
+  "kubernetes":{
+    "pod_name":"mypod-8f6bb798b-xxddw"
+  }
+}
+```
+
```
repeatedly: Could you move these text to `remove_keys` parameter section?

plugins/filter/record_transformer.md
```diff
@@ -59,6 +59,40 @@ into
 {"total":100, "count":10, "avg":"10"}
 ```
 
+This filter can be also used to remove fields through the `remove_keys`
+parameter (including nested keys, by means of `jsonpath` syntax)
```
repeatedly: > (including nested keys, by means of `jsonpath` syntax)

Remove this sentence. We don't say this syntax is `jsonpath` and `remove_keys` parameter section has a link to record accessor's syntax page.

kir4h:
Thanks for all of your work in `fluentd` @repeatedly!

Question: I was trying to use `record_modifier` for this same purpose (removing nested keys) and realized after a while this feature is only present in `record_transformer`? Since this page mentions `record_modifier`, should I add a disclaimer stating that if one needs to delete nested keys `record_transformer` should be used?

repeatedly:
> I add a disclaimer stating that if one needs to delete nested keys record_transformer should be used?

This seems good. Another solution is support same feature in record_modifier.

kir4h:
> > I add a disclaimer stating that if one needs to delete nested keys record_transformer should be used?
> 
> This seems good. Another solution is support same feature in record_modifier.

Seems better indeed. What I didn't like about adding the statement is that it might become false at some point if that behavior is added to `record_modifier` (there is no link between this documentation and that repository, so if someone implements it he won't be aware this documentation needs to be updated as well).

Two approaches I can think of for merging this PR:

1. Creating an issue in `record_modifier` repository and add a link to the issue in the documentation along with the statement
1. Not adding the statement

In any case I can give a try to the issue over the week-end. I don't have any ruby knowledge but having `record_transformer` PR as reference should help a lot.

kir4h:
@repeatedly do you want me to create the issue and add the statement, or should we merge this as it is now?

repeatedly:
Sorry for the delay. Just merged.

PR #170, opened 2020-04-22
update config examples to use out_forward instead of in_forward in service discovery 

sriramana:
configuration examples in service discovery examples use in_forward; update to out_forward


ganmacs:
You are right. Thank you.

PR #171, opened 2020-04-23
[1.0] Added more how-to examples

adnanrahic:
This PR adds another how-to example of sending syslog data to Sematext, which is an alternative for Splunk.

PR #172, opened 2020-04-24
[0.12] Added more how-to examples

adnanrahic:
This PR adds another how-to example for Fluentd version 0.12 of sending syslog data to Sematext, which is an alternative for Splunk. This is a sister-PR of #171 .


repeatedly:
New article is not needed for v0.12 because we, official repositories, stop all activities for v0.12.

adnanrahic:
Oh okay, thanks for letting me know. :ok_hand:

PR #174, opened 2020-05-08
Fix message length

marojor:
RFC-6587 defines the syslog frame as:

```
SYSLOG-FRAME = MSG-LEN SP SYSLOG-MSG
```

Thus the `MSG-LEN` should not include the space.


ganmacs:
Thanks!

PR #187, opened 2020-06-19
Add a bit more information about the time parameter / field

dobesv:
Note the appropriate format of the time field and add a warning about invalid time values.

### DIFF
--- a/plugins/input/http.md
+++ b/plugins/input/http.md
@@ -34,12 +34,17 @@ $ curl -X POST -d 'json={"foo":"bar"}' http://localhost:9880/app.log
 ```
 
 By default, timestamps are assigned to each record on arrival. You can
-override the timestamp using the `time` parameter.
+override the timestamp using the `time` parameter, either as a query
+parameter or in the payload.  It is an offset in seconds since Jan 1, 1970
+UTC.
 
 ```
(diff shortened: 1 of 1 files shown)


mlasevich:
That may be effectively correct, but misleading. Unless I am mistaken, query parameter and json field, while accomplishing similar thing of setting event time, are two distinct things handled by two different plugins. Query parameter is processed by the input plugin (`in_http`), while time field is processed by a parser parameter (`parser_json`)

repeatedly:
@dobesv Could you follow DCO? https://github.com/apps/dco

> while time field is processed by a parser parameter (parser_json)

With default mode, time field processing is handled in `in_http`. This is tricky for both single record and multiple records....

mlasevich:
> With default mode, time field processing is handled in `in_http`. This is tricky for both single record and multiple records....

Interesting, so there are 4 different/competing ways to provide time (and time syntax) in this case?
* query parameter in (`in_http`)
* field provided to the input (`in_http`)
* field provided to the parser (in any parser)
* `<extract>` clause in input (in any input)

is that correct?

dobesv:
> Could you follow DCO? https://github.com/apps/dco

I don't really understand what that is or how to apply it.  I think these docs are now obsolete anyway and have to be re-written, so this PR can be closed but used as "inspiration" for a better PR.

PR #192, opened 2020-07-10
Configuration Section: Language and grammar updates.

iamAzeem:
Configuration Section: Language and grammar updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,7 @@
 # Introduction
 
-![](.gitbook/assets/images/logo_documentation_1.0.png)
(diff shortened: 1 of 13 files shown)


repeatedly:
Thanks for great patch.

Could you remove TAB between parameter key and value, e.g. `@type             exec`?
This is unpopular in fluentd configuration.

iamAzeem:
@repeatedly: You're welcome!

> Could you remove TAB between parameter key and value, e.g. `@type exec`?
> This is unpopular in fluentd configuration.

Sure. I'll update all those patches.
I didn't know that it was not recommended practice.

Regards!

PR #202, opened 2020-07-15
Updates: Plugins - Buffer Section.

iamAzeem:
Updates: Plugins - Buffer Section.


repeatedly:
Thanks!

PR #203, opened 2020-07-15
Updates: Plugins - Storage.

iamAzeem:
Updates: Plugins - Storage.


repeatedly:
Thanks!

PR #204, opened 2020-07-15
Updates: Plugins - Service Discovery.

iamAzeem:
Updates: Plugins - Service Discovery.


repeatedly:
Thanks!

PR #212, opened 2020-07-18
Fixed broken links due to line breaks.

iamazeem:
Fixed broken links due to line breaks.


repeatedly:
Thanks!

PR #213, opened 2020-07-20
rpc require fluentd start not with --no-supervisor

like-inspur:
I found if config fluentd with --no-supervisor, rpc flunction will failed


repeatedly:
Thanks!

PR #214, opened 2020-07-22
Updates: Plugins - Parser. Fixed heading style.

iamazeem:
Updates: Plugins - Parser. Fixed heading style.


repeatedly:
Good catch!

PR #220, opened 2020-07-25
fix Environment Variable typo that changes PATH of configuration file(0.12)

### LINKED ISSUE #218: Enviromnent Vaiable name is wrong when changing configration file name
https://docs.fluentd.org/configuration/config-file#fluent_conf-environment-variable

> You can change the default configuration file location via FLUENT_CONF

x FLUENT_CONF
○ FLUENTD_CONF

It is FLUENTD_CONF in this code.
https://github.com/fluent/fluentd-docker-image/blob/master/entrypoint.sh.erb#L20

Anorlondo448:
Signed-off-by: Anorlondo448 <kanaend0605@gmail.com>

fix #218

### DIFF
--- a/configuration/config-file.md
+++ b/configuration/config-file.md
@@ -39,9 +39,9 @@ $ sudo vi /etc/fluent/fluent.conf
 
 #### FLUENT\_CONF environment variable
 
-You can change default configuration file location via `FLUENT_CONF`.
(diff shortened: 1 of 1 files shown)


ashie:
Current article is correct as described in https://github.com/fluent/fluentd-docs-gitbook/issues/218#issuecomment-664048465

PR #226, opened 2020-07-31
Updates: Image Descriptions

iamAzeem:
Updated description of images (screenshots and demos) from image filenames to adequate captions.
Removed description of images from plugin articles from right under their respective headings as they seem redundant.

Example: `tail.png` under the image of [tail input plugin](https://docs.fluentd.org/input/tail)


repeatedly:
Thanks!

PR #231, opened 2020-08-07
Updates: Reviewed the language and grammar changes.

iamazeem:
Updates: Reviewed the language and grammar changes.


iamAzeem:
@repeatedly: Could you please review and merge this or assign someone else?

developer/plugin-development.md
```diff
@@ -60,7 +60,7 @@ It is recommended to use the new v1 plugin API for writing new plugins.
 
 ### Send a patch or fork?
 
-If you have a problem with any existing plugins or new feature idea, sending a
+If you have a problem with any built-in plugins or a new feature idea, sending a
```
repeatedly, line 63: `built-in` means `fluentd` gem bundled plugins like `in_forward`. This section is for any plugins including 3rd party plugin, so I think `existing` is better.

plugins/output/forward.md
```diff
@@ -395,7 +395,7 @@ The client certificate path for TLS.
 |:------:|:-------:|:-------:|
 | string | nil     | 1.3.2   |
 
-The client private key path for TLS.
+The client 5private key path for TLS.
```
repeatedly, line 398: `5` seems typo

plugins/parser/multiline.md
```diff
@@ -23,20 +23,21 @@ See [Parse Section Configurations](/configuration/parse-section.md).
...
-| string | required parameter | 0.14.0  |
+| type   | default | version |
+|:-------|:--------|:--------|
+| string | `nil`   | 0.14.0  |
 
-Specifies the regexp patterns. For readability, you can separate regexp patterns
-into multiple `regexpN` parameters, See Rails Log example below. These patterns
-are joined and construct a regexp pattern with multiline mode.
+Specifies the regexp patterns. For readability, you can separate the regexp
+patterns into multiple `regexpN` parameters. See the Rails Log's example below.
```
repeatedly, line 38: This is original typo but `formatN` is correct, not `regexpN`

plugins/parser/multiline.md
```diff
@@ -23,20 +23,21 @@ See [Parse Section Configurations](/configuration/parse-section.md).
...
 
-If `format_firstline` is not specified, input plugin should store
-unmatched new lines in temporary buffer and try to match buffered logs
+If `format_firstline` is not specified, the input plugin should store the
+unmatched new lines in the temporary buffer and try to match the buffered logs
 with each new line.
 
 
-### `formatN`
+### `formatN` (required)
```
repeatedly, line 31: With this approach, link also contains `(required)`, right?
I want to avoid it. Because if we change such parameters to optional, the link will be broken.
Other same changes too.

repeatedly (commented):
Add comments

plugins/parser/multiline.md
iamAzeem: So, `regexpN` should be **patterns into multiple `formatN` parameters**, right?

repeatedly: Yes

plugins/parser/multiline.md
iamAzeem: I believe there was a mix of both conventions in the documentation. I tried to use only this one.
However, to avoid broken links, all the `(optional)` or `(required)` attributes should be moved with the description.
I'll update those as well.

Thank you for reviewing!

repeatedly:
Thanks for awesome work 👍

PR #244, opened 2020-10-13
buffer-section: update overflow_action block explanation

kenhys:
It seems that overflow_action block is used out-of-scope case, such
as avoiding high CPU usage. It is definitely wrong usage, but
known as a some sort of technique.


repeatedly:
thx!

PR #253, opened 2020-11-17
rewrite_tag_filter: Fix the incorrect pattern

harryge00:
The pattern should be `/^\[(\w+)\]/`. From the record `{"message":"[info]: ..."} `, the line starts with `[`, instead of the word `info`.

FYI: https://github.com/fluent/fluent-plugin-rewrite-tag-filter/issues/103


repeatedly:
Ah, good catch!

PR #258, opened 2020-12-02
output secondary_file: add howto to resend secondary file

kenhys:
It is useful to mention about how to recover from secondary file.


repeatedly:
Thanks!

PR #260, opened 2020-12-02
Improve elasticsearch output docs

ChrsMark:
This PR adds two minor documentation improvements:

- Improve wording about bulk indexing and point to ES bulk api docs
- Add link to buffer section on spot so as to have easier access to buffer options documentation.


repeatedly:
Thanks!

PR #261, opened 2020-12-09
configuration: fix typos about wUTC

kenhys:
wUTC ->
 UTC


repeatedly:
Sorry for the delay. We modified article structures / styles so this patch is voided. So fixed this issue via https://github.com/fluent/fluentd-docs-gitbook/commit/150f8d228eaae7eca21246005bc23f808a428c7c

PR #268, opened 2021-01-28
input/sample example - add tag

scharissis:
The example doesn't work as-is, as it's missing the required parameter 'tag'.


repeatedly:
Thanks!
Next, add sign by `-s` for DCO.

PR #280, opened 2021-03-09
Add explanation about time_format_fallbacks option

kenhys:


kenhys:
I'll tweak a bit.

kenhys:
Fixed paragraph to avoid misunderstanding about performance penalty.

cosmo0920 (approved):

PR #292, opened 2021-03-17
Add replacement config for formatter_ltsv

ashie:
It's introduced by https://github.com/fluent/fluentd/pull/3288

PR #295, opened 2021-03-30
Document new plugin features in v1.12.2

fujimotos:
I added descriptions for the features that we added in v1.12.2,
but have not yet documented.

With this merged, docs.fluentd.org should be keep in line with
the latest release. For details, see:

https://www.fluentd.org/blog/fluentd-v1.12.2-has-been-released

Signed-off-by: Fujimoto Seiji <fujimoto@ceptord.net>

### DIFF
--- a/formatter/ltsv.md
+++ b/formatter/ltsv.md
@@ -43,6 +43,14 @@ Adds `\n` to the result.
 
 Specify newline characters.
 
+### `replacement`
+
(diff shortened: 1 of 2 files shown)


ashie:
Oops, I wrongly added `replacement` to parser_ltsv at #292 ...
I'll revert it.

PR #303, opened 2021-04-13
1.0: clarify TLS configuration example

kenhys:
Before:

It seems that it is not clear that when certs generation is
enabled.

After:

It makes a clear explanation that pre-configured cert or auto-generated
certs are used or not.


fujimotos:
@kenhys  I merged this PR as ebcff9a7b68, with some minor fixes
(such as `a examples` -> `some examples`)

Since GitHub seems not to notice that this is already into master,
I close this PR manually.

PR #314, opened 2021-05-12
installation: Add a short introduction of td-agent

fujimotos:
`td-agent` is under-explained in the current guide, but it is easy to
understand in the context of Windows: It's a standalone package of
Fluentd that includes all the required dependencies.

Let's explain that, instead of telling users to read another article.

Signed-off-by: Fujimoto Seiji <fujimoto@ceptord.net>


ashie:
Could you resolve conflict?

fujimotos:
@ashie Conflict resolved.

ashie (approved):
LGTM

ashie:
Thanks!

PR #316, opened 2021-05-17
1.0: Document rotate_age and rotate_size in system config section

kenhys:
This configuration is mainly introduced to customize log rotation
settings on Windows (in a more comprehensive way).
Since implemented as system configuration,
this feature is supported on other platforms too.


ashie (approved):

kenhys:
Rebased with recent 1.0 branch.

PR #318, opened 2021-05-27
1.0: Mitigate descriptions for hardlink/symlink attack

cosmo0920:
This is usual software vulnerabilities for hardlink/symlink.
Fluentd sometimes uses symlink and predictable paths.
Predictable paths always are used in sigdump.
Which is known vulnerability of Fluentd agent.

Specifying `fs.protected_hardlinks` and `fs.protected_symlinks` to be set
up as `1` is popular approach to mitigate this vulnerability.

### DIFF
--- a/installation/before-install.md
+++ b/installation/before-install.md
@@ -57,6 +57,8 @@ These kernel options were originally taken from the presentation [How Netflix Tu
 
 ## Use sticky bit symlink/hardlink protection
 
+**NOTE:** CentOS 7 or later, Ubuntu 18.04 (bionic) or later, and Debian GNU/Linux 10 (buster) or later are supported these parameters.
+
 Fluentd sometimes uses predictable paths for dumping, writing files, and so on.
 For symlink attack protection, check the following parameters are set up as `1`:
 
@@ -67,7 +69,6 @@ fs.protected_symlinks = 1
 
 This settings are almost enough for time-of-check to time-of-use (TOCTOU, TOCTTOU or TOC/TOU) which is a class of software bugs.
 
-CentOS 7 or later, Ubuntu 18.04 (bionic) or later, and Debian GNU/Linux 10 (buster) or later are supported these parameters.
 If you turned off these protection, please turn on them.
 
 Use `sysctl -p` command or reboot your node for the changes to take effect.
--- a/installation/before-install.md
+++ b/installation/before-install.md
@@ -60,6 +60,8 @@ These kernel options were originally taken from the presentation [How Netflix Tu
 **NOTE:** CentOS 7 or later, Ubuntu 18.04 (bionic) or later, and Debian GNU/Linux 10 (buster) or later are supported these parameters.
 
 Fluentd sometimes uses predictable paths for dumping, writing files, and so on.
+This default settings for the protections are in `/etc/sysctl.d/10-link-restrictions.conf`, or `/usr/lib/sysctl.d/50-default.conf` or elsewhere.
+
 For symlink attack protection, check the following parameters are set up as `1`:
 
 ```text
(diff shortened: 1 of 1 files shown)


cosmo0920:
Note that TOC/TOU protection settings is already enabled by default on Ubuntu.

installation/before-install.md
```diff
@@ -55,5 +55,18 @@ Use `sysctl -p` command or reboot your node for the changes to take effect.
...
+Fluentd sometimes uses predictable paths for dumping, writing files, and so on.
+For symlink attack protection, check the following parameters are set up as `1`:
+
+```text
+fs.protected_hardlinks = 1
+fs.protected_symlinks = 1
+```
 
+This settings are almost enough for time-of-check to time-of-use (TOCTOU, TOCTTOU or TOC/TOU) which is a class of software bugs.
+
```
kenhys, line 69: It may be better to explain who should customize these parameters.
In the recent version of Linux distribution, it seems already enabled.
(At least Debian 10 (buster) or later and Ubuntu 20.04 or later, these configuration is enabled)

kenhys: Addition, it may be useful if the link to the reference to these parameters.

https://www.kernel.org/doc/Documentation/sysctl/fs.txt

cosmo0920: > (At least Debian 10 (buster) or later and Ubuntu 20.04 or later, these configuration is enabled)

I'm still stay in Ubuntu 18.04. It is also enabled these parameters.

cosmo0920: I added a notice for sticky bit protections.

And I noticed that CentOS 6 does not support these parameters.... Oh my gosh.... :confounded:

installation/before-install.md
```diff
@@ -55,5 +55,21 @@ Use `sysctl -p` command or reboot your node for the changes to take effect.
...
+CentOS 7 or later, Ubuntu 18.04 (bionic) or later, and Debian GNU/Linux 10 (buster) or later are supported these parameters.
+If you truned off these protection, please turn on them.
```
kenhys, line 71: typo? turned off

cosmo0920: I'd fixed them in https://github.com/fluent/fluentd-docs-gitbook/pull/318/commits/a1c3d075161434eaa7c09232202dada0daf22a1f and https://github.com/fluent/fluentd-docs-gitbook/pull/318/commits/c7e652e2c46004c86d9978b1bf288ae3385eb483.

kenhys (approved):
LGTM.

installation/before-install.md
```diff
@@ -55,5 +55,22 @@ Use `sysctl -p` command or reboot your node for the changes to take effect.
 
 These kernel options were originally taken from the presentation [How Netflix Tunes EC2 Instances for Performance](https://www.slideshare.net/brendangregg/how-netflix-tunes-ec2-instances-for-performance) by [Brendan Gregg](http://www.brendangregg.com/), Senior Performance Architect at AWS re:Invent 2017.
 
-If this article is incorrect or outdated, or omits critical information, please [let us know](https://github.com/fluent/fluentd-docs-gitbook/issues?state=open). [Fluentd](http://www.fluentd.org/) is an open-source project under [Cloud Native Computing Foundation \(CNCF\)](https://cncf.io/). All components are available under the Apache 2 License.
+## Use sticky bit symlink/hardlink protection
+
+**NOTE:** CentOS 7 or later, Ubuntu 18.04 (bionic) or later, and Debian GNU/Linux 10 (buster) or later are supported these parameters.
+
+Fluentd sometimes uses predictable paths for dumping, writing files, and so on.
+For symlink attack protection, check the following parameters are set up as `1`:
```
ashie, line 63: Should be described which file have to be modified (`/etc/sysctl.conf`).
Although it can be predictable from the previous section, should be clarified as same as `sysctl -p`.

cosmo0920: Ah, we should. Thanks for the pointing out.

cosmo0920: Fixed in https://github.com/fluent/fluentd-docs-gitbook/pull/318/commits/318417e0d71158e3514699b7f28339df36343dad.

ashie (approved):

cosmo0920:
Thanks!

PR #319, opened 2021-05-27
Fix: Change hostname to host

juergen-walter:
Without the fix, Fluentd does not start and logs an error
...
`error_class=Fluent::ConfigError error="'host' parameter is required, in section server"`
...
The change adjusts to what is in the current master in the linked example
https://github.com/kzk/fluentd-prometheus-config-example/blob/c03d7797606c63830ebd6715d223ede282f19e34/fluentd.conf#L28

### DIFF
--- a/monitoring-fluentd/monitoring-prometheus.md
+++ b/monitoring-fluentd/monitoring-prometheus.md
@@ -70,7 +70,7 @@ Configure the `copy` plugin with `prometheus` output plugin to count the outgoin
     @type forward
     <server>
       name myserver1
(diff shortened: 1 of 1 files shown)


repeatedly:
Thanks for the patch.
But this change doesn't fix the problem because you change `prometheus` output, not `forward` output.
Could you recheck the patch?

juergen-walter:
Thank you for carefully reviewing it. Force pushed the fix.

ashie (approved):
LGTM
Thanks!

PR #322, opened 2021-06-02
1.0: use {% include %} directive for footer content

kenhys:
If we need to modify footer content in each page, we should also
modify ~180 page files.
It seems better to split it into asset file.

For example, there is a case that footer content is displayed in a PR, with `{% %}` directive, it can make it smaller diff.
https://github.com/fluent/fluentd-docs-gitbook/pull/318/files


ashie:
Hmm, footers are gone.
I'll revert it.

ashie:
Sorry, I didn't check it locally because my local gitbook-cli is currently broken...

kenhys:
image

```
% gitbook --version
CLI version: 2.3.2
GitBook version: 3.2.3
```

Hmm, in my local environment, it works as expected.

kenhys:
I've created the issue for following up. https://github.com/fluent/fluentd-docs-gitbook/issues/324

PR #328, opened 2021-06-08
1.0: add blocking_reload_interval explanation

kenhys:
Add explanation about https://github.com/fluent/fluentd/pull/3396


kenhys:
need to reconsider it.

Issue #325, opened 2021-06-04
roundrobin output plugin example references tcp output plugin, which does not exist in core fluentd


LorenKeagle:
I spent a long time trying to figure out how to send my events over a raw TCP socket.  Fluentd core has a raw TCP input plugin, but no raw TCP output plugin that I can find. 

However, the documentation for this roundrobin output plugin suggests that core fluentd has a tcp output plugin that can be used.  Since this is not the case, I recommend changing the example to use a core output plugin.

Or better yet, include a TCP output plugin in the core implementation :-)

kenhys:
> I recommend changing the example to use a core output plugin.

It is a good suggestion not to make new user confused.

kenhys:
It should be fixed by https://github.com/fluent/fluentd-docs-gitbook/pull/329

kenhys:
#329 was merged :+1:

PR #329, opened 2021-06-08
1.0: use core plugin for roundrobin example

kenhys:
Using non standard output plugin makes a new user confused.

Closes: #325

### DIFF
--- a/output/roundrobin.md
+++ b/output/roundrobin.md
@@ -11,20 +11,20 @@ It is included in Fluentd's core.
   @type roundrobin
 
   <store>
-    @type tcp
-    host 192.168.1.21
+    @type http
+    endpoint http://192.168.1.21
     weight 3
     # ...
   </store>
   <store>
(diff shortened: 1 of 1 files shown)


output/roundrobin.md
```diff
@@ -11,20 +11,32 @@ It is included in Fluentd's core.
   @type roundrobin
 
   <store>
-    @type tcp
-    host 192.168.1.21
+    @type http
+    endpoint http://192.168.1.21
+    <format>
```
ashie, line 16: `format` and ` json_array` aren't needed, because they aren't mandatory options.
Since this page is a document for roundrobin plugin, information that is not related to it should be suppressed as much as possible.

kenhys: It should be, thanks!

kenhys:
Removed not mandatory options.

ashie (approved):
LGTM

PR #330, opened 2021-06-19
in_http: add use_204_response

nokute78:
I added missing `use_204_response`.

- Fluentd Pull Request: 
  - https://github.com/fluent/fluentd/pull/2640
- Commit: 
  - https://github.com/fluent/fluentd/commit/ee2ae7b0178ec1f13e53ff18d35e7b5818bc55b1
  - It means the option is supported from v1.8.0
- Note:
  - The option will be deprecated at v2. See also https://github.com/fluent/fluentd/pull/2640#issuecomment-539439936

### DIFF
--- a/input/http.md
+++ b/input/http.md
@@ -144,6 +144,14 @@ Example:
 
 Responds with an empty GIF image of 1x1 pixel \(rather than an empty string\).
 
+### `use_204_response`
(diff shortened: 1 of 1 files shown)


ashie:
Thanks!

Issue #334, opened 2021-06-22
Explain reserve_time


Guillaume-Mayer:
I have a hard time understanding the **reserve_time** option of the parser filter.
If I enable or disable the option, I cannot see any difference.
Also it seems slightly different of **keep_time_key** but not sure how.

The documentation is very succinct and show no examples.
If this could be improved it would be nice.

Thanks

kenhys:
I hope that it will help https://github.com/fluent/fluentd-docs-gitbook/pull/335

kenhys:
#335 wes merged.

Guillaume-Mayer:
That's perfectly explained! Thanks!

PR #335, opened 2021-06-24
Explain reserve_time with an execution result

kenhys:
Closes: #334

It will help to understand the behavior of reesrve_time.


ashie (approved):
LGTM

PR #337, opened 2021-06-29
in_tail: Revise read_from_head

### LINKED ISSUE #336: Document behavior of `read_from_head true` for existing logs when pos file is also configured
**Is your feature request related to a problem? Please describe.**
There are multiple bugs (for example, https://github.com/fluent/fluentd/issues/1595) where people are confused about `in_tail` behavior with `read_from_head true` AND pos file configured. My testing shows that when `read_from_head true` AND pos file configured the existing logs are NOT read from head, but instead read from the last pos file position. If this is in fact the expected behavior, please document it at https://docs.fluentd.org/input/tail#read_from_head.

**Describe the solution you'd like**
My testing shows that when `read_from_head true` AND pos file configured the existing logs are NOT read from head, but instead read from the last pos file position. If this is in fact the expected behavior, please document it at https://docs.fluentd.org/input/tail#read_from_head.

ashie:
* Describe correct start position, it should be last read position if `pos_fileP is configured. Fix #336
* Revise other explanations to make easy to understand merits & demerits of this option.

### DIFF
--- a/input/tail.md
+++ b/input/tail.md
@@ -174,7 +174,7 @@ Skips the refresh of the watch list on startup. This reduces the startup time wh
 | :--- | :--- | :--- |
 | bool | false | 0.14.0 |
 
-Starts to read the logs from the head of the file, not tail.
+Starts to read the logs from the head of the file or the last read position recorded in `pos_file`, not tail.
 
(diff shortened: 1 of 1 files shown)


kenhys (approved):
LGTM.

PR #347, opened 2021-08-02
1.0: Add metrics plugin descriptions

cosmo0920:
Add documentations for metrics plugin.

Mainly for https://github.com/fluent/fluentd/pull/3479, https://github.com/fluent/fluentd/pull/3473, and https://github.com/fluent/fluentd/pull/3440

### DIFF
--- a/metrics/README.md
+++ b/metrics/README.md
@@ -32,6 +32,11 @@ Here is an example with `metrics_local`:
 </system>
 ```
 
+`local` type plugin should provide equivalent behavior before Fluentd v1.13.
+This metrics type should provide single numeric value storing functionality.
+
+And this `local` type plugin should be used by default.
+
 ## List of Built-in Metrics Plugins
 
 * [`local`](local.md)
--- a/metrics/README.md
+++ b/metrics/README.md
@@ -1,0 +1,53 @@
+# Metrics Plugins
+
+Fluentd has nine \(9\) types of plugins:
+
+* [Input](../input/)
+* [Parser](../parser/)
+* [Filter](../filter/)
+* [Output](../output/)
+* [Formatter](../formatter/)
+* [Storage](../storage)
+* [Service Discovery](../service_discovery/)
+* [Buffer](../buffer/)
+* [Metrics](./)
+
+This article gives an overview of Metrics Plugin.
+
+## Overview
+
+Sometimes, the input/filter/output plugin needs to save its internal metrics in memory, influxdb or prometheus format ready in instances. Fluentd has a pluggable system called Metrics that lets a plugin store and reuse its internal state as metrics instances.
+
+## How To Use
+
+On Fluentd core, metrics plugin will handled on `<metrics>` on `<system>` to set up easily.
+
+Here is an example with `metrics_local`:
+
+```text
+<system>
+  <metrics>
+    @type local
+  </metrics>
+</system>
+```
+
+## List of Built-in Metrics Plugins
+
+* [`local`](local.md)
+
+## List of Base Plugin classes with Metrics support
+
+* `Fluent::Plugin::Input` for Input plugin base class
+* `Fluent::Plugin::Output` for most of output plugin base class
+* `Fluent::Plugin::Filter` for Filter plugin base class
+* `Fluent::Plugin::MultiOutput` for [out_copy](../output/copy.md) plugin base class
+* `Fluent::Plugin::BareOutput` for fluent-plugin-forest output plugin base class
+
+## List of 3rd party metrics plugins
+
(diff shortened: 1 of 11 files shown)


metrics/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+```text
+<system>
+  <metrics>
+    @type local
+  </metrics>
+</system>
+```
+
+## List of Built-in Metrics
```
ashie: Please add `Plugins`

```diff
- ## List of Built-in Metrics
+ ## List of Built-in Metrics Plugins
```

ashie (changes requested):

metrics/README.md
cosmo0920: Fixed in https://github.com/fluent/fluentd-docs-gitbook/pull/347/commits/f57d9aa2fac954406f6477338dbb85c7a086b82f.

ashie (approved):

cosmo0920:
This PR should be merged into 1.0 after Fluentd v1.14 is shipped.

metrics/README.md
```diff
@@ -0,0 +1,53 @@
...
+* [Parser](../parser/)
+* [Filter](../filter/)
+* [Output](../output/)
+* [Formatter](../formatter/)
+* [Storage](../storage)
+* [Service Discovery](../service_discovery/)
+* [Buffer](../buffer/)
+* [Metrics](./)
+
+This article gives an overview of Metrics Plugin.
```
kenhys, line 15: It may be off-topic, should we list all types here? (I felt a redundant) :thinking:

kenhys: Currently every plugin, top of them, we list it.

metrics/local.md
```diff
@@ -0,0 +1,29 @@
...
+With this configuration:
+
+```text
+<system>
+  <metrics>
+    @type local
+  </metrics>
+</system>
+```
+
```
kenhys, line 26: It seems that Fluentd's user may be confused what can do with bult-in `@type local`? It is not explained well.
It says that "The `local` metrics plugin stores the values on memory." but how to access it?

It may be covered with fluent-plugin-metrics-cmetrics, but it may be unfriendly.

metrics/local.md
```diff
@@ -0,0 +1,29 @@
...
+The above configuration will save the internal metrics for plugins on memory. As a result, you can retrive metrics from memory and also you can replace with your custom metrics plugin.
+
```
kenhys, line 28: So, It is important to explain `@type local` behavior here to clarify the difference.

kenhys (commented):
IMHO, from user point of view, it needs more explanation "how to".

metrics/local.md
cosmo0920: Yeah, we should do. 
Actually, `@type local` metrics plugin has equivalent functionality for previous single value based Ruby instance variables.
This type of plugin is just used for storing numerical values on memory.

metrics/local.md
cosmo0920: No, built-in `local` type plugin should be covering the previous usages for storing single numerical values.
This behavior will be changed by other 3rd party plugins but built-in plugin should have equivalent functionality before Fluentd v1.13 series do.

metrics/README.md
cosmo0920: This should be working as pointer or providing overview of Fluentd's plugin type.
I would like to leave as-is for now.
I think that removing this plugin type list is unfriendly for newbies.

metrics/local.md
cosmo0920: And I think that this question should be solved by adding fluent-plugin-metrics-cmetrics' plugin descriptions.
This plugin has very interesting functionality, which is native prometheus/influxdb format support.
These native supported format will handle for more rich labeled prometheus metrics and fast format translations among cmetrics, influxdb, and prometheus' counter/gauge formats.
This interesting concepts and usage will be revealed within this month. 😛

metrics/local.md
cosmo0920: Fixed in https://github.com/fluent/fluentd-docs-gitbook/pull/347/commits/db265fa15d5101cea7cb8b36d184eaa062d0afc9.

metrics/local.md
kenhys: >  I think that this question should be solved by adding fluent-plugin-metrics-cmetrics' plugin descriptions.

If so, it seems ok.

kenhys (approved):
LGTM

kenhys:
merged :congratulations:

cosmo0920:
Hooray! :tada:

PR #352, opened 2021-08-21
fix 'path' parameter style

nokute78:
https://docs.fluentd.org/output/elasticsearch#path-optional

I modified a style of 'path' parameter.
I added missing backquotes.


ashie:
Thanks!

PR #353, opened 2021-08-23
Add entry about Debian 11 (bullseye)

kenhys:
When https://toolbelt.treasuredata.com/sh/install-debian-bullseye-td-agent4.sh is published, we should merge this PR.


ashie:
> When https://toolbelt.treasuredata.com/sh/install-debian-bullseye-td-agent4.sh is published, we should merge this PR.

Done!

PR #363, opened 2021-09-24
Fix apiversion in daemonset example

sibucan:
The extensions/v1beta1 apiversion has been removed from Kubernetes since v1.17


cosmo0920 (approved):
Good catch! :+1:

Issue #360, opened 2021-09-08
Documentation calls a backward slash a forward slash


gmcelhoe:
### Describe the bug

The text of the section on escaping characters in config files shows the glyph of a backward slash, but the text erroneously names it a forward slash.

### Link to the problematic documentation

https://docs.fluentd.org/configuration/config-file#in-double-quoted-string-literal-is-the-escape-character

### Expected explanation

The character \  should be referred to as a backward slash.

kenhys:
@gmcelhoe could you send PR, please?

kenhys:
Fixed by #364

PR #364, opened 2021-09-27
Use appropriate *backward* slash

kenhys:
'\\' should be backward slash, not forward slash.

Closes: #360

Reported by @gmcelhoe

Signed-off-by: Kentaro Hayashi <hayashi@clear-code.com>

### DIFF
--- a/configuration/config-file.md
+++ b/configuration/config-file.md
@@ -605,7 +605,7 @@ The `config-xxx` mixins use `"${}"`, not `"#{}"`. These embedded configurations 
 
 ### In double-quoted string literal, `\` is the escape character
 
-The forward slash `\` is interpreted as an escape character. You need `\` for setting `"`, `\r`, `\n`, `\t`, `\` or several characters in double-quoted string literal.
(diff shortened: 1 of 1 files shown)


configuration/config-file.md
```diff
@@ -605,7 +605,7 @@ The `config-xxx` mixins use `"${}"`, not `"#{}"`. These embedded configurations
 
 ### In double-quoted string literal, `\` is the escape character
 
-The forward slash `\` is interpreted as an escape character. You need `\` for setting `"`, `\r`, `\n`, `\t`, `\` or several characters in double-quoted string literal.
+The backward slash `\` is interpreted as an escape character. You need `\` for setting `"`, `\r`, `\n`, `\t`, `\` or several characters in double-quoted string literal.
```
ashie: Not bad but probably `backslask` is more familiar than `backward slash`.

ashie (approved):

configuration/config-file.md
kenhys: Thanks, fixed now.

PR #367, opened 2021-10-05
1.0: Fix sample logs at install section

mishina2228:
* Fix a command to tail the log of calyptia-fluentd.
* Fix markdown.


kenhys:
@mishina2228 
Good catch!

cosmo0920:
Thanks!

PR #368, opened 2021-10-06
Use curl -fsSL -o- consistently

brsolomon-deloitte:
Add some flags to curl to raise error on failed request and ensure
output is stdout. Use bash rather than text blocks.

### DIFF
--- a/installation/install-by-deb.md
+++ b/installation/install-by-deb.md
@@ -32,64 +32,64 @@ A shell script is provided to automate the installation process for each version
 
 For Ubuntu Focal:
 
-```text
+```bash
 # td-agent 4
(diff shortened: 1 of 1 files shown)


kenhys:
It seems to use `-fsSL` is better, but is `-o-` really needed? (I'm not sure)

brsolomon-deloitte:
Yes, I suppose it is redundant in this case. From `man curl`

> Specifying the output as '-' (a single dash) will  force the output to be done to stdout.

But that would really only be useful if multiple `-o` are passed, now that I think about it.

brsolomon-deloitte:
> It seems to use `-fsSL` is better, but is `-o-` really needed? (I'm not sure)

Removed in b2c2f06.

kenhys:
@brsolomon-deloitte 

Thanks!

PR #376, opened 2021-10-20
1.0: in_tail_with_throttle: Add `<group>` directive description

Pranjal-Gupta2:
Documentation for [fluent/fluentd#3535](https://github.com/fluent/fluentd/pull/3535)

Signed-off-by: Pranjal-Gupta2 <pranjal.gupta2@ibm.com>


cosmo0920 (changes requested):
Could you rewrite to be corresponding to https://github.com/fluent/fluentd/pull/3535?
https://github.com/fluent/fluentd/pull/3535 is very reasonable for us but the documentation is also needed to ship into stable release.

cosmo0920 (approved):
Seems good for me.

input/tail.md
```diff
@@ -402,6 +412,80 @@ The `@log_level` option allows the user to set different levels of logging for e
 
 Refer to the [Logging](../deployment/logging.md) for more details.
 
+
+### `<group>` Section \(highly recommended\)
```
daipom:
```suggestion
### `<group>` Section
```

I think this function is rather optional.
(Use when we want more detailed control than `read_bytes_limit_per_second`)

input/tail.md
```diff
@@ -17,6 +17,16 @@ It is included in Fluentd's core.
...
+  <group>
+    rate_period 30s
+    <rule>
+      match {
+          namespace: /space1|space2|space2/,
+          podname: /app.*/,
+      }
+      limit 200
+    </rule>
+  </group>
```
daipom: For the same reason, I think it would be better to leave the simple configuration example here as it is
and have a separate configuration example for this feature.

daipom: How about moving this example to just below the `<group>` Section and reverting this block?

input/tail.md
ashie: Removed in c0a9c1e35276d9e33ec8280b6a33f6aa9c46c253

daipom:
Thank you for the fix!

input/tail.md
ashie: Just removed in c0a9c1e35276d9e33ec8280b6a33f6aa9c46c253
Since another example already exists just under `<group>`, I simply removed it.

ashie:
I merge this since v1.15 has been released.
Thank you for your effort!

PR #384, opened 2021-12-19
fix: repo link DaemonSet preset settings to v1.14

ffuentes:


ffuentes:
Hello, maybe is better to the docker-image folder to avoid fixed version reference?

kenhys:
@ffuentes Thanks!

PR #385, opened 2021-12-22
out_forward, out_file: Add missing default value of <buffer>

ashie:
See also: https://github.com/fluent/fluentd/issues/3564

Signed-off-by: Takuro Ashie <ashie@clear-code.com>


cosmo0920 (approved):
God catch! :+1:

PR #386, opened 2021-12-27
Rewrite wrong description about `path` of file buffer plugin

ashie:
Although it seems based on v0.12, it's not applicable for v0.14 or later.
We should rewrite it.

refs: https://github.com/fluent/fluentd/issues/3552


ashie:
Actual rewriting has been done by @daipom at #415, it looks good to me.

daipom:
Thanks for reviewing and merging!

PR #390, opened 2022-02-14
1.0: Add OpenSearch plugin documentation

cosmo0920:

### DIFF
--- a/output/opensearch.md
+++ b/output/opensearch.md
@@ -1,6 +1,6 @@
 # opensearch
 
-The `out_opensearch` Output plugin writes records into Opensearch. By default, it creates records using [bulk api](https://opensearch.org/docs/latest/opensearch/rest-api/document-apis/bulk/) which performs multiple indexing operations in a single API call. This reduces overhead and can greatly increase indexing speed. This means that when you first import records using the plugin, records are not immediately pushed to OpenSearch.
+The `out_opensearch` Output plugin writes records into OpenSearch. By default, it creates records using [bulk api](https://opensearch.org/docs/latest/opensearch/rest-api/document-apis/bulk/) which performs multiple indexing operations in a single API call. This reduces overhead and can greatly increase indexing speed. This means that when you first import records using the plugin, records are not immediately pushed to OpenSearch.
 
 Records will be sent to OpenSearch when the `chunk_keys` condition has been met. To change the output frequency, please specify the `time` in `chunk_keys` and specify `timekey` value in the configuration.
 
--- a/output/opensearch.md
+++ b/output/opensearch.md
@@ -18,7 +18,7 @@ $ fluent-gem install fluent-plugin-opensearch
 
 ## Example Configuration
 
-Here is a simple working configuration which should serve as a good starting point for most users:
+Here is a simple working configuration for OpenSearch instance that is running on localhost:
 
 ```text
 <match my.logs>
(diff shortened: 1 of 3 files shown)


output/opensearch.md
```diff
@@ -0,0 +1,163 @@
...
+
+The `out_opensearch` Output plugin writes records into Opensearch. By default, it creates records using [bulk api](https://opensearch.org/docs/latest/opensearch/rest-api/document-apis/bulk/) which performs multiple indexing operations in a single API call. This reduces overhead and can greatly increase indexing speed. This means that when you first import records using the plugin, records are not immediately pushed to OpenSearch.
+
+Records will be sent to OpenSearch when the `chunk_keys` condition has been met. To change the output frequency, please specify the `time` in `chunk_keys` and specify `timekey` value in the configuration.
+
+This document does not describe all the parameters. For details, refer to the **Further Reading** section.
+
+## Installation
+
+Since `out_opensearch` has been included in the standard distribution of `td-agent` since v4.4.0 and the alternative distribution of `calyptia-fluentd` since v1.3.4, `td-agent`/`calyptia-fluentd` users do not need to install it manually.
```
cosmo0920: But, I'm not sure the exact version of OpenSearch plugin included version of td-agent.

output/opensearch.md
```diff
@@ -0,0 +1,163 @@
+# opensearch
+
+The `out_opensearch` Output plugin writes records into Opensearch. By default, it creates records using [bulk api](https://opensearch.org/docs/latest/opensearch/rest-api/document-apis/bulk/) which performs multiple indexing operations in a single API call. This reduces overhead and can greatly increase indexing speed. This means that when you first import records using the plugin, records are not immediately pushed to OpenSearch.
```
kenhys:
```suggestion
The `out_opensearch` Output plugin writes records into OpenSearch. By default, it creates records using [bulk api](https://opensearch.org/docs/latest/opensearch/rest-api/document-apis/bulk/) which performs multiple indexing operations in a single API call. This reduces overhead and can greatly increase indexing speed. This means that when you first import records using the plugin, records are not immediately pushed to OpenSearch.
```

output/opensearch.md
```diff
@@ -0,0 +1,163 @@
...
+If you have installed Fluentd without `td-agent` or `calyptia-fluentd`, please install this plugin using `fluent-gem`:
+
+```text
+$ fluent-gem install fluent-plugin-opensearch
+```
+
+## Example Configuration
+
+Here is a simple working configuration which should serve as a good starting point for most users:
+
```
kenhys, line 26: It may be obvious, but it is kind to note "for OpenSearch instance on localhost" or something.

kenhys (commented):
I've added a few comments.

output/opensearch.md
cosmo0920: Fixed in https://github.com/fluent/fluentd-docs-gitbook/pull/390/commits/798582a31dffd7c5dc2a71ccc267e4ac37ced8e2.

output/opensearch.md
agup006: It would be good to merge prior to being available in td-agent, and then adding once available

cosmo0920: Sure. I fixed it in https://github.com/fluent/fluentd-docs-gitbook/pull/390/commits/76ef4607b1fe4620d8b4b12e82553109746f51f6.

cosmo0920:
@kenhys @ashie Can we merge this OpenSearch plugin descriptions? It seems that OpenSearch plugin gets initial users increasing momentum. We should follow this momentum and want to merge this doc into official docs soon.

output/opensearch.md
ashie: FYI:
Probably the next version of td-agent (which includes opensearch plugin) will be v4.3.1 but not determined yet.

The minor version of td-agent is tied with the one of fluentd.
When the minor version of fluentd is incremeneted, td-agent's one is also incremented.

ashie (approved):

cosmo0920:
Thanks!

PR #395, opened 2022-03-25
1.0: Fix wrong description of total retry time

daipom:
For https://github.com/fluent/fluentd/pull/3649


ashie (approved):
LGTM

daipom:
I found another wrong description, so fixed it.
In addition, I feel we need examples of retry timings, so I added some.

PR #392, opened 2022-02-25
1.0: describe new option `linger_timeout` of transport section

daipom:
Doc for https://github.com/fluent/fluentd/pull/3644.

PR #398, opened 2022-04-04
Add description to `linger_timeout` behavior about Windows

daipom:
This is corresponding to https://github.com/fluent/fluentd/pull/3701

This option is added in https://github.com/fluent/fluentd/pull/3644
and the description is added in https://github.com/fluent/fluentd-docs-gitbook/pull/392

There was a lack of consideration regarding the default behavior on Windows.

This fixes it.


daipom:
Thanks for merging!

PR #407, opened 2022-06-15
1.0: Add restart_worker_interval description

daipom:
This is for the new parameter restart_worker_interval that will be added in the fix:

* https://github.com/fluent/fluentd/pull/3768


ashie (approved):

ashie:
v1.15 has been released.
Thanks!

PR #409, opened 2022-06-29
Fix inaccuracy in UDP input parse section

### LINKED ISSUE #408: Inaccuracy in parser section of udp input
### Describe the bug

In the [`<parse>` section](https://docs.fluentd.org/input/udp#less-than-parse-greater-than-section) of the UDP input, the term "`in_tcp`" should be replaced with the term "`in_udp`".

> `in_tcp` uses the parser plugin to parse the payload. [...]

Should read like this:

> `in_udp` uses the parser plugin to parse the payload. [...]

<https://docs.fluentd.org/input/udp#less-than-parse-greater-than-section>

### Expected explanation

Documentation should be accurate.

### Additional context

_No response_

dtvez:
This merge request fixes the inaccuracy in the UDP input parse section (fixes #408).

### DIFF
--- a/input/udp.md
+++ b/input/udp.md
@@ -141,7 +141,7 @@ Removes the newline from the end of the incoming payload.
 | :--- | :--- | :--- |
 | true | false | 0.14.10 |
 
-`in_tcp` uses the parser plugin to parse the payload.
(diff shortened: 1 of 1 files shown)


ashie:
Thanks!

PR #411, opened 2022-07-03
Fix wrong setting and typo

daipom:
https://docs.fluentd.org/configuration/config-file-yaml#share-the-same-parameters

I fixed some points here that I felt were not correct.


ashie:
Thanks!

PR #412, opened 2022-07-03
Fix wrong settings

daipom:
https://docs.fluentd.org/configuration/config-file-yaml

There were a few examples of settings that were not right, so I fixed them.


ashie:
Thanks!

PR #413, opened 2022-07-04
Add multiple workers config for both traditional and yaml

daipom:
I know the description about multiple workers is here.

* https://docs.fluentd.org/deployment/multi-process-workers

However, I need the YAML config examples for multiple workers.

So I want to add it to https://docs.fluentd.org/configuration/config-file-yaml.

Since the contents of that page link to the contents of https://docs.fluentd.org/configuration/config-file,
I add it to both of them.


ashie:
Thanks!

daipom:
Thanks for merging!

PR #414, opened 2022-07-04
Fix a broken internal link

daipom:
I'm sorry. The fix of https://github.com/fluent/fluentd-docs-gitbook/pull/410 is inadequate.

The `label` link in https://docs.fluentd.org/configuration/config-file#common-plugin-parameters isn't still working after all.

Somehow it worked in my local environment, but it doesn't work on the production environment.
It is because `.` is not removed from DOM-id on the production environment.

### DIFF
--- a/configuration/config-file.md
+++ b/configuration/config-file.md
@@ -507,7 +507,7 @@ These parameters are reserved and are prefixed with an `@` symbol:
 
 * `@label`: specifies the label symbol. See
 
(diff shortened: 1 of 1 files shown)


kenhys:
LGTM. merged.

daipom:
Thanks for merging!

daipom:
I have confirmed this works correctly on the product environment.

PR #415, opened 2022-07-07
Add fix to "Rewrite wrong description about path of file buffer plugin"

daipom:
This is for https://github.com/fluent/fluentd-docs-gitbook/pull/386 .

The `version` of `path` parameter is still TBD.

Should we change from the original version number `0.9.0` to the version number with the current specification?


ashie:
Thanks a lot!

> Should we change from the original version number 0.9.0 to the version number with the current specification?

I'll check the history...

daipom:
Thanks!

PR #416, opened 2022-07-08
Use yaml syntax

daipom:
How about using yaml syntax for code-blocks in the yaml-configuration page?

* https://docs.fluentd.org/configuration/config-file-yaml

Screenshot from 2022-07-08 10-25-01


ashie (approved):

ashie:
Thanks!

daipom:
Thanks for merging!

Issue #417, opened 2022-07-12
"Optimize the Network Kernel Parameters" setting may cause fluentd to fail to restart.


migimigi:
### Describe the bug

Sometimes when restarting fluentd I get the error "Address already in use - bind(2) for 0.0.0.0:24224".

### Link to the problematic documentation

https://docs.fluentd.org/installation/before-install#optimize-the-network-kernel-parameters

### Expected explanation

Explicitly reserve the ports used by fluentd using net.ipv4.ip_local_reserved_ports.

### Additional context

I searched for binding failures and found a blog post about a similar situation.
[Fluentd failed to start because of “Address already in use — bind(2) for 0.0.0.0:24224”](https://komamitsu.medium.com/fluentd-failed-to-start-because-of-address-already-in-use-bind-2-for-0-0-0-0-24224-def62042bffe)
In this article, the author discovered that if the server socket port is in the ephemeral port range, the client socket may use that port as its src port.
As a solution, it is suggested to narrow the range of ephemeral ports, but this will reduce the number of available ports, so instead use net.ipv4.ip_local_reserved_ports to reserve the ports used by fluentd.

migimigi:
Merged!

PR #418, opened 2022-07-12
Add net.ipv4.ip_local_reserved_ports to /etc/sysctl.conf to prevent s…

migimigi:
#417
In the Optimize the Network Kernel Parameters section, I added an additional setting to /etc/sysctl.conf with net.ipv4.ip_local_reserved_ports to explicitly reserve the ports to be used by fluentd.


ashie:
Thanks!

PR #424, opened 2022-09-05
Remove a wrong description about overflow_action of out_forward

ashie:
out_forward never send error to client. It just doesn't send ack
response on error.
See also: https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1#response

Fix https://github.com/fluent/fluentd/issues/3742

### DIFF
--- a/output/README.md
+++ b/output/README.md
@@ -163,7 +163,7 @@ Supported modes:
 
 * `throw_exception` \(default\)
 
-  This mode throws the`BufferOverflowError` exception to the input plugin. How `BufferOverflowError` is handled depends on the input plugins, e.g. tail input stops reading new lines, forward input returns an error to forward output. This action is suitable for streaming.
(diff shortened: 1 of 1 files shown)


daipom (commented):
Thanks for the fix!

> out_forward never send error to client. It just doesn't send ack
response on error.

in_forward?
I think the server is in_forward and the client is out_forward.

Is there any reason why the description about forwarding should be completely removed instead of fixing it as follows?

```diff
- forward input returns an error to forward output
+ forward input returns no ack response to forward output
```

ashie:
> > out_forward never send error to client. It just doesn't send ack
> > response on error.
> 
> in_forward? I think the server is in_forward and the client is out_forward.

Sorry, it was typo.

> Is there any reason why the description about forwarding should be completely removed instead of fixing it as follows?
> 
> ```diff
> - forward input returns an error to forward output
> + forward input returns no ack response to forward output
> ```

Because in_forward doesn't send ack response by default, it sends it only when `require_ack_response true` is set. I think describing it here is too complicated as an example of `overflow_action`.

daipom (approved):
I see! Thanks!
I'm fine with this fix!

ashie:
Thanks for your review!

PR #425, opened 2022-09-05
monitor_agent: add notification about sensitive information

kenhys:
Generally, monitor agent will not be in public, but it is better
to notify a concern about sensitive data.


ashie (approved):
LGTM

ashie:
Thanks!

PR #429, opened 2022-09-27
install-by-msi: add content about unregistering td-agent v3 before uninstalling

daipom:
Some users had trouble with this problem and were unable to install v4.
So I felt the need to add this kind of documentation.


ashie (approved):
LGTM

ashie:
Thanks!

daipom:
Thanks for merging!

PR #430, opened 2022-09-27
Fix docs about log rotation

daipom:
I would like to add 2 points about log ratation with this fix.

* Reflect in `deployment/logging.md` that we can set log rotation options with `system-config` after v1.13.0.
* Although those 2 options have a default value, we need to specify at least one of those options explicitly to enable log rotation.

At first, I thought the default values were wrong, but those default values are managed on ServerEngine side. 

* `rotate_age`
  * https://docs.fluentd.org/deployment/system-config#rotate_age
  * https://github.com/fluent/fluentd/blob/v1.15.2/lib/fluent/system_config.rb#L63
* `rotate_size`
  * https://docs.fluentd.org/deployment/system-config#rotate_size
  * https://github.com/fluent/fluentd/blob/v1.15.2/lib/fluent/system_config.rb#L76

In addition, we need to specify at least one of those options explicitly.

* https://github.com/fluent/fluentd/blob/v1.15.2/lib/fluent/supervisor.rb#L543-L549

This is complicated and hard to understand, so I think we need a document like this.

### DIFF
--- a/deployment/logging.md
+++ b/deployment/logging.md
@@ -156,7 +156,25 @@ $ fluentd -o /path/to/log_file
 
 ### Log Rotation Setting
 
-By default, Fluentd does not rotate log files. You can configure this behavior via command-line options:
+By default, Fluentd does not rotate log files. You can configure this behavior via system-config after v1.13.0.
+
+It can be configured through `<log>` directive under `<system>`:
+
+```text
+<system>
+  <log>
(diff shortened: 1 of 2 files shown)


ashie (approved):
LGTM

ashie:
Thanks!

daipom:
Thanks for merging!

PR #433, opened 2022-10-12
Another failure reason for unknown placeholder

honnix:
I think missing `key` in the input record is another reason for unknown placeholder error after reading https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/filter_record_transformer.rb, because when that happens we don't "prepare" it.

### DIFF
--- a/filter/record_transformer.md
+++ b/filter/record_transformer.md
@@ -296,7 +296,7 @@ ${record.dig("top", "nest1", "nest2")}
 
 ### I got `unknown placeholder ${record['msg']} found` error, why?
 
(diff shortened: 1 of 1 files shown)


honnix:
This probably gives a better answer to https://github.com/fluent/fluentd/issues/2530

kenhys:
Even though v1.16.3, it will be reproducible with something like:

`enable_ruby false` and nonexistent `404` key:

```
<filter dummy>
  @type record_transformer
  #enable_ruby
  <record>
   nokey ${record['404']}
  </record>
</filter>
```


```
2023-12-06 18:45:44 +0900 [warn]: #0 unknown placeholder `${record['404']}` found
```

kenhys (approved):
LGTM.

PR #435, opened 2022-11-09
Add missing require statement for tests

camposer:


ashie:
If a plugin is generated by `fluent-plugin-generate` command,  it will be loaded via generated helper.rb.
But the sample code doesn't use helper.rb so that `fluent/test` should be loaded explicitly by the test code.
Thanks!

PR #445, opened 2023-03-06
out_secondary_file: add missing option explanation

daipom:
This explanation is missing since the beginning.

* https://github.com/fluent/fluentd/blob/master/CHANGELOG.md#release-v0145---20160906
* https://github.com/fluent/fluentd/pull/1154

### DIFF
--- a/output/secondary_file.md
+++ b/output/secondary_file.md
@@ -99,6 +99,16 @@ This makes parallel file processing easy. But if you want to disable this behavi
 dump.bin
 ```
 
+### `compress`
+
+| type | default | version |
+| :--- | :--- | :--- |
+| enum | text | 1.0.0 |
+
+Supported compress types: `text`, `gzip`
+
+When `gzip` is specified, `.gz` is automatically added to the output file path as its suffix.
+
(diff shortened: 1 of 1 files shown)


ashie (approved):
LGTM

daipom:
Thanks for merging!

output/secondary_file.md
```diff
@@ -99,6 +99,16 @@ This makes parallel file processing easy. But if you want to disable this behavi
 dump.bin
 ```
 
+### `compress`
+
+| type | default | version |
+| :--- | :--- | :--- |
+| enum | text | 1.0.0 |
```
ashie, line 106: Actual version which adds this feature seems 0.14.5 but it's meaning less to specify exact version.
* https://github.com/fluent/fluentd/blob/master/CHANGELOG.md#release-v0145---20160906
* https://github.com/fluent/fluentd/pull/1154

daipom: Yes, the other options are the same.
Or maybe every option here could be changed to v0.14.5.

PR #447, opened 2023-03-29
in_tcp: Fix wrong supported version for "send_keepalive_packet"

ashie:
Follow up for #438


daipom (approved):
LGTM. Thanks for fixing this!

PR #449, opened 2023-04-06
Plugin development: Update test helper template

daipom:
Document Update for

* https://github.com/fluent/fluentd/pull/4128


daipom:
I am also concerned about the following content.

https://github.com/fluent/fluentd-docs-gitbook/blob/b9ba13a60bec45c9da728cd09706e23bc5981643/plugin-development/plugin-update-from-v0.12.md?plain=1#L633-L643

This may need to be updated as well, but we should consider it separately from this PR.

ashie:
Thanks!

daipom:
Thanks for your review!

PR #451, opened 2023-04-18
0.12: improve expression

### LINKED ISSUE #450: Better name for the "Monitoring Prometheus" page
### Describe the bug

The name "Monitoring Prometheus" in https://docs.fluentd.org/v/0.12/articles/monitoring-prometheus makes it sound like it's about using Fluentd to monitor Prometheus.

It would be more descriptive if this page was renamed to something like "Monitoring with Prometheus"

https://docs.fluentd.org/v/0.12/articles/monitoring-prometheus

### Expected explanation

Rename to "Monitoring with Prometheus"

### Additional context

_No response_

daipom:
* Fix #450

I used `by` because on the current `1.0` page, the titles are `Monitoring by Prometheus` and `Monitoring by REST API`.

I didn't change `Api` to `API` because `Api` is entirely used on `0.12`.


ashie (approved):
LGTM

daipom:
Thanks!

PR #456, opened 2023-07-06
installation: markup separated command line example

kenhys:
It makes easy to copy & paste.


daipom (approved):
Thanks! LGTM!

Should we remove the example of td-agent 3?

PR #457, opened 2023-07-07
Split deprecated v3 installation

kenhys:
As already announced EOL in Dec 2021,  let's remove it.

https://www.fluentd.org/blog/schedule-for-td-agent-3-eol


ashie:
How about move it to a separated page instead of removing it completely.
Some users might still want to refer it (since some our customers still use it).

kenhys:
Thanks, I'll fix it.

kenhys:
Instead of removing them, keep it separated section like this:

image

and navigation to v3.

image

also added guide for migration from v3.

image

ashie (approved):

ashie:
Thanks!

PR #458, opened 2023-07-14
installation: Improve explanation about td-agent v3 deb package

daipom:
There are explanations about packages other than td-agent.
We don't need it on this page.


kenhys (approved):
LGTM.

kenhys:
good catch!

daipom:
Thanks for your review!

PR #472, opened 2023-09-08
output/forward: add notification about specific use case

kenhys:
Use case:

```
   tls_cert_path /etc/fluent/cert/fluentd.crt
   <server>
     host 192.168.10.100
     port 24224
   </server>
```

If CN in fluentd.crt is 192.168.10.100 (IP address), internally the value of hostname is nil, so #verify_hostname raise undefined method ascii_only? exception. To avoid exception, we must suppress verification with tls_veriry_hostname false.

See https://github.com/fluent/fluentd/discussions/4289

### DIFF
--- a/output/forward.md
+++ b/output/forward.md
@@ -333,6 +333,11 @@ Allows self-signed certificates or not.
 
 Verifies hostname of servers and certificates or not in TLS transport.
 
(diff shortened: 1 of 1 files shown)


daipom (approved):
Thanks so much!

PR #478, opened 2023-10-16
input/http: Added more details about the Content-Type header

DavidRomao:
Hello, 
A couple weeks ago I stumbled upon the issue described here https://github.com/fluent/fluentd/issues/4293 
It turned that there was no bug on Fluentd but the docs weren't very clear about the Content-Type which needs to be used. 

I wasn't aware of the default behavoir of `curl` when the Content Type wasn't specified, which led to some confusion.  So I tried to clarify that part. 
I've also added another entry under Troubleshooting documenting the behaviour I experienced.

I propose the changes in this PR in an attempt to include some more information in hope that it the Fluentd behaviour easier to understand.

Thanks! 
Any feedback is welcome If there's anything that could be improved further.

### DIFF
--- a/input/http.md
+++ b/input/http.md
@@ -22,12 +22,14 @@ For the full list of the configurable options, see the [Parameters](http.md#para
 
 ## Basic Usage
 
-Here is a simple example to post a record using `curl`.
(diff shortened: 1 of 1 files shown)


kenhys (approved):
LGTM.

Thanks!

PR #480, opened 2023-11-29
fix: :bug: `force_path_style` is deprecated

mehdiMj-ir:
This option is deprecated and it show up while I was restarting fluentd service [warn]: #0 'force_path_style' parameter is deprecated: S3 will drop path style API in 2020: See https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

Signed-off-by: mehdiMj <mehdi.is@live.com>


daipom (changes requested):
Thanks!
That option is deprecated in `out_s3`, so we should remove it from the document.

* https://github.com/fluent/fluent-plugin-s3/pull/274

Could you add DCO to your commit?

https://github.com/fluent/fluentd-docs-gitbook/pull/480/checks?check_run_id=19137000625

mehdiMj-ir:
Thank you for considering my pull request, I add DCO to my commit message.

daipom (approved):
LGTM. Thanks!

PR #484, opened 2023-12-19
fluent-cat: update options for v1.16.3

daipom:
The options in https://docs.fluentd.org/deployment/command-line-option#fluent-cat are old.
This updates them to the latest version, Fluentd v1.16.3.


ashie (approved):

daipom:
Thanks!

PR #485, opened 2024-01-02
Remove warning about issue that was closed years ago

ecerulm:
The current documentation at https://docs.fluentd.org/input/tail#pos_file-highly-recommended says 

> in_tail removes the untracked file position at startup. It means that the content of pos_file keeps growing until a restart when you tails lots of files with the dynamic path setting.
> This [issue](https://github.com/fluent/fluentd/issues/1126) will be fixed in future.

but that issue was closed 2020-02-30. 

So I think the documentation should remove the reference to that closed issue or alternatively it could be replaced with 

> `in_tail` removes the untracked file position at startup. It means that the content of `pos_file` keeps growing until a restart when you tails lots of files with the dynamic path setting. **This issue can be solved by using `pos_file_compaction_interval`**

### DIFF
--- a/input/tail.md
+++ b/input/tail.md
@@ -269,9 +269,10 @@ pos_file /var/log/td-agent/tmp/access.log.pos
 
 Don't share `pos_file` between `in_tail` configurations. It causes unexpected behavior e.g. corrupt `pos_file` content.
 
-`in_tail` removes the untracked file position at startup. It means that the content of `pos_file` keeps growing until a restart when you tails lots of files with the dynamic path setting.
-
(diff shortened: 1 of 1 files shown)


kenhys:
Thank you feedback.

It seems that the latter explanation may be better.
Could you change so that?

P.S. Contributing documentation is very welcome, but it is required to add signed commit. Could you follow [DCO](https://github.com/fluent/fluentd-docs-gitbook/pull/485/checks?check_run_id=20091313650)?

ecerulm:
@kenhys, I added the DCO / Signed-off , and change the text.

kenhys (approved):
LGTM.

kenhys:
Thank you for contributing

PR #486, opened 2024-02-05
Update status about EOL of td-agent v4

kenhys:
See https://www.fluentd.org/blog/schedule-for-td-agent-4-eol


kenhys:
Rebased with 1.0

kenhys:
Try to use hints.

ashie (approved):

PR #487, opened 2024-02-05
Enable hints plugin by default

kenhys:
It aims to markup emphasized topics (info,tip,danger,working).

See https://www.npmjs.com/package/gitbook-plugin-hints


kenhys:
Here is the example which is locally tested with honkit.

image

Not fully tested with deployed version, so if it is not compatible well, it should be reverted later.

ashie (approved):

PR #490, opened 2024-03-26
Markup copy&paste friendly

kenhys:
* Use bash highlighting ($ is added by default)
* Eliminate "fluent-package 5" and "fluent-package 5 (LTS)" label from code blocks


kenhys:
Changed to:

image

daipom (approved):
LGTM. Thanks! Nice idea!

PR #493, opened 2024-04-11
Add note that in_sample does not support parser

kenhys:


configuration/parse-section.md
```diff
@@ -61,18 +61,18 @@ Here's the list of built-in parser plugins:
 
 The default value of the following parameters will be overridden by the individual parser plugins:
 
+{% hint style='info' %}
+NOTE: parse parameters will not work with `@type sample` input plugin.
+{% endhint %}
```
daipom: Thanks for this fix!

Could you tell me why we need this note?

> NOTE: parse parameters will not work with `@type sample` input plugin.

daipom: Is this related to https://github.com/fluent/fluentd/discussions/4461?

daipom: I don't understand this discussion in detail, but my first impression is that this is an issue of Input plugin and should be described in https://docs.fluentd.org/input/sample.

I'm sorry, but if there is something I don't understand, please let me know.

kenhys: Yes. not confused for that case.

kenhys: Fixed it.

configuration/parse-section.md
```diff
@@ -59,20 +59,20 @@ Here's the list of built-in parser plugins:
 
 ### Parse Parameters
 
+{% hint style='info' %}
+NOTE: It depends on parser implementation whether plugin supports the following parse parameters. See each plugin's documentation in detail.
```
cosmo0920:
```suggestion
NOTE: It depends on parser implementation whether a plugin supports the following parse parameters. See each plugin's documentation in detail.
```

configuration/parse-section.md
```diff
@@ -59,20 +59,20 @@ Here's the list of built-in parser plugins:
...
+{% endhint %}
+
 The default value of the following parameters will be overridden by the individual parser plugins:
 
 * **`types`** \(hash\) \(optional\): Specify types for converting field into another
 
-  type. See below "The detail of types parameter" section.
+  type. See below the detail of "types Parameter" section.
```
cosmo0920:
```suggestion
  type. See below the details of the "types Parameter" section.
```

cosmo0920 (commented):
It would be good to apply the suggestions to clarify the meanings.

kenhys:
Thanks!

configuration/parse-section.md
```diff
@@ -59,20 +59,20 @@ Here's the list of built-in parser plugins:
 
 ### Parse Parameters
 
+{% hint style='info' %}
+NOTE: It depends on parser implementation whether a plugin supports the following parse parameters. See each plugin's documentation in detail.
```
daipom: Hmm, I thought we were trying to say that some input plugins do not support the `<parse>` section, so be careful.
Such as follows.

```suggestion
NOTE: It depends on a plugin (Input plugin in many cases) whether the plugin supports the `<parse>` section. See each plugin's documentation in detail.
```

The current sentences can be read as follows for me.
If my understanding is somehow incorrect, please let me know.

> It depends on the parser plugin whether the following parameters are supported.

cosmo0920: It may not wrong. So, probably, we need to clarify this sentence with more concrete description.

```suggestion
NOTE: It depends on an input plugin's capabilities whether can handle a `<parser>` section or not. See each plugin's documentation in detail.
```

daipom: Thanks! I agree with the suggestion of @cosmo0920

kenhys:
rebased with 1.0 branch.

daipom (approved):
Thanks!

daipom:
I failed to include `singed-off` and `co-authored-by` into the squashed commit message. I'm sorry @kenhys @cosmo0920  :cry:

PR #494, opened 2024-04-15
yaml: Add description for log_level per plugin

Shingo-Nakayama:
>https://github.com/fluent/fluentd/issues/4463

In this issue, we found out that YAML format doesn't support the `$log_level`.
So I would like to add the situation and workaround for this issue.


kenhys (approved):
LGTM.

PR #499, opened 2024-05-09
config: yaml: Add the Parse setting in Format Tips section

Shingo-Nakayama:
I had trouble setting up parse in YAML format because there was no explanation of how to do it.
So I want to add the method of parse setting in YAML format to Format Tips section.

### DIFF
--- a/configuration/config-file-yaml.md
+++ b/configuration/config-file-yaml.md
@@ -552,4 +552,18 @@ The backslash `\` is interpreted as an escape character. You need `\` for settin
 str_param:   "foo\nbar" # \n is interpreted as actual LF character
 ```
 
+### Parse setting
+You can use `parse:` to set up parse.
+Example: in_tail plugin
+```yaml
+config:
+  - source:
+      $type: tail
+      tag: sample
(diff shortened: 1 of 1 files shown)


configuration/config-file-yaml.md
```diff
@@ -552,4 +552,18 @@ The backslash `\` is interpreted as an escape character. You need `\` for settin
 str_param:   "foo\nbar" # \n is interpreted as actual LF character
 ```
 
+### Parse setting
+You can use `parse:` to set up parse.
+Example: in_tail plugin
+```yaml
```
daipom: Thanks! Basically looks good!

How about this?

```suggestion
### Parse setting

You can use `parse:` to set up the parser for the input plugin.

Example: in_tail plugin

```yaml
```

daipom: Before:

Screenshot from 2024-05-10 13-58-06

After:

Screenshot from 2024-05-10 14-04-44

daipom (approved):
LGTM! Thanks!

PR #502, opened 2024-05-15
in_http: add descriptions about Content-Type

daipom:
https://docs.fluentd.org/input/http

Need to add `application/x-ndjson` and `application/csp-report`.

Related:

* https://github.com/fluent/fluentd/pull/3616
* https://github.com/fluent/fluentd/pull/4282

### DIFF
--- a/input/http.md
+++ b/input/http.md
@@ -51,8 +51,8 @@ msgpack=`echo -e "msgpack=\x81\xa3foo\xa3bar"`
 curl -X POST -d "$msgpack" http://localhost:9880/app.log
 ```
 
-Some `Content-Type` other than `application/x-www-form-urlencoded` support specific formats.
-In that case, the format type specification as `json=` in the data is not necessary.
+Some Media Types other than `application/x-www-form-urlencoded` support specific formats.
+If those Media Types are specified with `Content-Type: `, the prefix such as `json=` is not necessary for posting data.
 
 Example: Post JSON data with `Content-Type: application/json`:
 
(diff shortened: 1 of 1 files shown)


input/http.md
```diff
@@ -22,11 +22,43 @@ For the full list of the configurable options, see the [Parameters](http.md#para
...
+
+Example: post MessagePack data with the tag "app.log":
+
+```bash
+msgpack=`echo -e "msgpack=\x81\xa3foo\xa3bar"`
+curl -X POST -d "$msgpack" http://localhost:9880/app.log
+```
+
+Some `Content-Type` other than `application/x-www-form-urlencoded` support specific formats.
+In that case, the format type specification as `json=` in the data is not necessary.
```
kenhys: > In that case, the format type specification as `json=` in the data is not necessary.

How about another wording something like: 

```
If media type is specified with `Content-Type: `, then content type prefix such as `json=` is not necessary for posting data.

Example: ....
```

kenhys (changes requested):

input/http.md
daipom: Thanks for your review!
Could you tell me which points are your concerns?
I guess the aim is to make the explanation more concrete. Is this right?
If so, how about something like this?

```diff
- Some `Content-Type` other than `application/x-www-form-urlencoded` support specific formats.
- In that case, the format type specification as `json=` in the data is not necessary.
+ Some Media Types other than `application/x-www-form-urlencoded` support specific formats.
+ If those Media Types are specified with `Content-Type: `, the prefix such as `json=` is not necessary for posting data.
```

daipom: > content type prefix such as `json=`

I have a concern about whether we can describe the prefix as `content type prefix`.
Isn't this prefix for the data format for parsing, rather than for `Content-Type`?

kenhys: > I guess the aim is to make the explanation more concrete. Is this right?

Yes. so https://github.com/fluent/fluentd-docs-gitbook/pull/502#discussion_r1630703860 suggested one seems good.

daipom: OK! Thanks! I will fix it!

daipom: Fixed!

kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #503, opened 2024-05-31
Copy missing LICENSE file into 1.0 branch too

kenhys:
LICENSE file was already commit into master branch, but as it is not default branch, it may be confused what license is applied to all components.

The documentation of footer contains mention about Apache 2.0 license, but there are some content files which lacks that notification. (It should be also fixed too in another PR)

Above situation should be solved by put LICENSE file into 1.0 branch too.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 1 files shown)


daipom (approved):
LGTM.
Thanks for this.

PR #507, opened 2024-06-20
plugin Development:README:Make it more easily to read

Shingo-Nakayama:
As in the [https://github.com/fluent/fluentd/discussions/4511](url), the documentation was confusing about file names when creating custom plugins.
So I made the following modifications:
- Move the original OVERVIEW to the back as Note
- In the new OVERVIEW, I wrote that the file name need to be the TIME prefix.

### DIFF
--- a/plugin-development/README.md
+++ b/plugin-development/README.md
@@ -1,16 +1,9 @@
 # Plugin Development
 
-## Installing Custom Plugins
+## Overview
+With fluentd, you can install and create custom plugins.
 
-To install a plugin, put the Ruby script in `/etc/fluent/plugin` directory.
-
-Alternatively, you can create a Ruby Gem package that includes:
-
-```text
-lib/fluent/plugin/<TYPE>_<NAME>.rb
-```
-
-The `TYPE` prefix could be:
+To install or create a custom plugin, the file name need to be `<TYPE>_<NAME>.rb`. Note that `TYPE` is a prefix indicating the type of plugin, and could be:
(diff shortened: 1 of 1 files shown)


plugin-development/README.md
```diff
@@ -1,16 +1,9 @@
...
-
-To install a plugin, put the Ruby script in `/etc/fluent/plugin` directory.
-
-Alternatively, you can create a Ruby Gem package that includes:
-
-```text
-lib/fluent/plugin/<TYPE>_<NAME>.rb
-```
+## Overview
+With fluentd, you can install and create custom plugins.
```
daipom:
```suggestion
## Overview

With fluentd, you can install and create custom plugins.
```

plugin-development/README.md
```diff
@@ -23,6 +16,15 @@ The `TYPE` prefix could be:
 | `buf` | buffer |
 | `sd` | service discovery |
 
+## Installing Custom Plugins
+
+To install a plugin, put the Ruby script in `/etc/fluent/plugin` directory.(The file name need to `TYPE` prefix.)
```
daipom:
```suggestion
To install a plugin, put the Ruby script in `/etc/fluent/plugin` directory. (The file name needs to have the `TYPE` prefix.)
```

daipom (changes requested):
Thanks!

daipom (approved):
LGTM. Thanks!

PR #508, opened 2024-07-08
rpm: add alert about dropped support

kenhys:


kenhys:
We need raise an attention.

https://docs.fluentd.org/installation/install-by-rpm#step-1-install-from-rpm-repository

image

installation/install-by-rpm.md
```diff
@@ -57,6 +57,10 @@ Executing this script will automatically install `fluent-package` on your machin
 
 We use `$releasever` for repository path in the script and `$releasever` should be the major version only like `"7"`. If your environment uses some other format like `"7.2"`, change it to the major version only or set up .repo file manually.
 
+{% hint style='danger' %}
+Since v5.0.4, RHEL 7 / CentOS 7 was not supported anymore because CentOS 7 had reached EOL (June, 2024). Please consider migrating to the other release.
```
daipom: I'm not sure, but it looks more natural to me to focus on the current status (tense).

```suggestion
Since v5.0.4, RHEL 7 / CentOS 7 is not supported anymore because CentOS 7 has reached EOL (June, 2024). Please consider migrating to another release.
```

installation/install-by-rpm.md
```diff
@@ -57,6 +57,10 @@ Executing this script will automatically install `fluent-package` on your machin
 
 We use `$releasever` for repository path in the script and `$releasever` should be the major version only like `"7"`. If your environment uses some other format like `"7.2"`, change it to the major version only or set up .repo file manually.
```
daipom: Update the example version.

```suggestion
We use `$releasever` for repository path in the script and `$releasever` should be the major version only like `"9"`. If your environment uses some other format like `"9.2"`, change it to the major version only or set up .repo file manually.
```

daipom (approved):
Thanks! LGTM!

I have commented on some minor points.

installation/install-by-rpm.md
kenhys: Thanks, it should be.

PR #509, opened 2024-07-24
rpm deb: add upgrade note about v3

kenhys:
Directly upgrade workflow is not supported.
We should not it.


kenhys:
It will be rendered.

image

daipom (approved):
LGTM! Thanks!

PR #511, opened 2024-08-07
faq: update supported Ruby version

Watson1978:
According to https://github.com/fluent/fluentd/blob/5618dd0e86bf0bc47a6bed2da50e37943ac2c8e9/fluentd.gemspec#L23,
Fluentd has supported Ruby 2.7 or later.


daipom (approved):
LGTM. Thanks!

PR #512, opened 2024-08-08
Fix typos

Watson1978:


output/copy.md
```diff
@@ -111,7 +111,7 @@ This parameter is deprecated since v1.8.1. Use `copy_mode` instead.
 
 `out_copy` shares a record between `store` plugins by default.
 
-If `true`, `out_copy` passes dupped record to each `store` plugin. This behavior is similar to `copy_mode shallow`.
+If `true`, `out_copy` passes dumped record to each `store` plugin. This behavior is similar to `copy_mode shallow`.
```
kenhys: duplicated? (or intended dup in Ruby context)

daipom: Looks like this is not a typo.
I think `dupped` wants to say Ruby's `dup`.

Watson1978: Ah. I see. Thanks

daipom (approved):
LGTM. Thanks!

PR #513, opened 2024-08-09
Add note about RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR

kenhys:
For example, RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR=0.9 causes launching time penalty:

* ruby-3.3.4(YJIT) 0.9 ~40 secs
* ruby-3.2.5(YJIT) 0.9 ~3 secs
* ruby-3.1.6(YJIT) 0.9 ~1 secs
* ruby-3.0.7 0.9 ~1 secs

See https://github.com/fluent/fluentd/issues/4545

### DIFF
--- a/deployment/performance-tuning-single-process.md
+++ b/deployment/performance-tuning-single-process.md
@@ -74,7 +74,11 @@ Here's a quote from the documentation:
 > Do full GC when the number of old objects is more than R * N
 > where R is this factor and N is the number of old objects just after last full GC.
 
-So, the default GC behavior does not call full GC until the number of old objects reaches `2.0 * before old objects`. This improves the throughput but it grows the total memory usage. This setting is not good for the low resource environment e.g. a small container. For such cases, try `RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR=0.9` or `RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR=1.2`.
+So, the default GC behavior does not call full GC until the number of old objects reaches `2.0 * before old objects`. This improves the throughput but it grows the total memory usage. This setting is not good for the low resource environment e.g. a small container. For such cases, try `RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR=1.2`.
+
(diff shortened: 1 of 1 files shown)


kenhys:
Add explanation like this:

image

deployment/performance-tuning-single-process.md
```diff
@@ -76,6 +76,11 @@ Here's a quote from the documentation:
 
 So, the default GC behavior does not call full GC until the number of old objects reaches `2.0 * before old objects`. This improves the throughput but it grows the total memory usage. This setting is not good for the low resource environment e.g. a small container. For such cases, try `RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR=0.9` or `RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR=1.2`.
 
+{% hint style='danger' %}
+There is a case that `RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR=0.9` causes a delay of launching Fluentd.
+As excessively tune to reduce memory usage causes performance degrade, configure `RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR` parameter by your own risk.
+{% endhint %}
+
```
daipom: How about more aggressively preventing the use of values less than 1.0?

```suggestion
So, the default GC behavior does not call full GC until the number of old objects reaches `2.0 * before old objects`. This improves the throughput but it grows the total memory usage. This setting is not good for the low resource environment e.g. a small container. For such cases, try `RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR=1.2`, etc.

{% hint style='danger' %}
Do not set a value lower than `1.0` unless there is a special reason, because there are cases where setting a value lower than `1.0` causes a delay of launching Fluentd.
{% endhint %}
```

kenhys: Thanks, a bit modified it.

daipom (approved):
Thanks! LGTM!

PR #515, opened 2024-08-19
sample: add explanation about reuse_record

kenhys:


kenhys:
I've forgot to assign reviewer.

daipom (approved):
LGTM! It is very clear to me! Thanks!

PR #521, opened 2024-12-03
transport-section: add ensure_fips option

kenhys:


configuration/transport-section.md
```diff
@@ -91,6 +91,10 @@ The max size of socket receive buffer for TCP/UDP. This is used in `SO_RCVBUF` s
   * OpenSSL 1.0.0 or higher default.
 * `insecure` \[bool\]
   * Default: `false` \(uses secure connection with `tls`\)
+* `ensure_fips`: \[bool\]
+  * Default: `false`
+  * Specifies whether it must use FIPS mode with OpenSSL. If `true`, Fluentd will check FIPS mode is supported in your environment, if not, just aborts.
```
daipom: Thanks!
How about adding version info?

```suggestion
  * Default: `false`
  * Version: 1.18.0
  * Specifies whether it must use FIPS mode with OpenSSL. If `true`, Fluentd will check FIPS mode is supported in your environment, if not, just aborts.
```

kenhys: It looks good. Thanks!

daipom (approved):
Thanks!

PR #523, opened 2024-12-23
linux-capability: add note for using CAP_DAC_READ_SEARCH in docker container

Watson1978:
Related to https://github.com/fluent/fluentd/issues/4213

### DIFF
--- a/deployment/linux-capability.md
+++ b/deployment/linux-capability.md
@@ -181,3 +181,19 @@ Fluentd, which is running by a non-root user, does not complain with `Permission
 
 If this article is incorrect or outdated, or omits critical information, please [let us know](https://github.com/fluent/fluentd-docs-gitbook/issues?state=open). [Fluentd](http://www.fluentd.org/) is an open-source project under [Cloud Native Computing Foundation \(CNCF\)](https://cncf.io/). All components are available under the Apache 2 License.
 
+## Capability handling on docker container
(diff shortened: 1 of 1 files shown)


daipom (commented):
Thanks so much for this enhancement!
It looks basically good to me, but how about moving this section as follows?

I feel a separate section would be easier to understand because this is not directly related to `in_tail`.

Screenshot from 2024-12-23 18-08-58

deployment/linux-capability.md
```diff
@@ -84,6 +84,21 @@ Inheritable: dac_override, dac_read_search
 Permitted:   dac_override, dac_read_search
 ```
 
+#### Note: using CAP\_DAC\_READ\_SEARCH in docker container
+
+When using `CAP_DAC_READ_SEARCH` in a Docker container, you need to add the `--cap-add DAC_READ_SEARCH` option to the `docker run` command.
+Or, if you are using `docker compose`, you need to add `cap_add` to the service definition.
```
kenhys: It might be better to explain "the context" - such as collecting logs with non-root user.

kenhys (changes requested):

Watson1978:
Thanks. 
I moved this to last section and add some explain.

kenhys (approved):
LGTM.

daipom (approved):
LGTM! Thanks!

PR #524, opened 2024-12-26
Bump required version to Ruby 3.2

kenhys:
Since Fluentd v1.19.0 was changed to require Ruby 3.2.


kenhys:
This PR is reserved for future release of Fluentd, so still drant and labeled as pending.

daipom (approved):

PR #527, opened 2025-01-10
source-only mode: add initial document

daipom:
https://github.com/fluent/fluentd/pull/4661


deployment/source-only-mode.md
```diff
@@ -0,0 +1,76 @@
...
+This file buffer is the buffer of [out_buffer](../output/buffer.md).
+It works with the default settings of `out_buffer` except for the following points:
+
+* `path` is automatically determined by default.
+* `overflow_action` is `drop_oldest_chunk` by default.
+
+If need, you can configure some options for the buffer in System Configuration.
+Please see the following for details.
+
+* [System Configuration - \<source_only_buffer\> section](system-config.md#less-than-source_only_buffer-greater-than-section).
```
Watson1978: Looks for me that the `less-than-source_only_buffer-greater-than-section` anchor is missing.
So, the browser cannot navigate to the appropriate part of the page...

Watson1978: When I preview this in local environment, it shows as following

image

Seems the unexpected message is shown as link.

deployment/source-only-mode.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Recovering
+
+If Fluentd stops with the temporary buffer left, you need to recover the buffer to launch Fluentd with source-only mode again.
+
+Note that a different path will be used each time unless you configure the temporary buffer path explicitly.
+In this case, you can recover the buffer as follows.
+
+1. Configure the left buffer path explicitly.
```
Watson1978:
```suggestion
1. Configure the remaining buffer path explicitly.
```

deployment/source-only-mode.md
```diff
@@ -0,0 +1,76 @@
...
+1. Start Fluentd with source-only mode again.
+1. Send `SIGWINCH` to the supervisor to load the buffer.
+
+If this recovering is necessary, i.e., Fluentd stops with the temporary buffer left, the following warning log will be output.
```
Watson1978:
```suggestion
If this recovery is necessary, i.e., Fluentd stops with the temporary buffer in place, the following warning log will be output.
```

deployment/source-only-mode.md
```diff
@@ -0,0 +1,76 @@
...
+* `overflow_action` is `drop_oldest_chunk` by default.
+
+If need, you can configure some options for the buffer in System Configuration.
+Please see the following for details.
+
+* [System Configuration - \<source_only_buffer\> section](system-config.md#less-than-source_only_buffer-greater-than-section).
+
+### Recovering
+
+If Fluentd stops with the temporary buffer left, you need to recover the buffer to launch Fluentd with source-only mode again.
```
Watson1978:
```suggestion
If Fluentd stops with the temporary buffer in place, you need to recover the buffer to launch Fluentd with source-only mode again.
```

deployment/system-config.md
```diff
@@ -100,6 +100,20 @@ Logs event verbosely.
...
+| :--- | :--- | :--- |
+| bool | nil | 1.18.0 |
+
+Not supported on Windows.
+
+Invokes a fluentd only with input plugins.
+
+See [Source Only Mode](source-only-mode.md) for details.
+
+See also [\<source_only_buffer\> section](#less-than-source_only_buffer-greater-than-section).
```
Watson1978: Looks for me that the less-than-source_only_buffer-greater-than-section anchor is missing.
So, the browser cannot navigate to the appropriate part of the page...

Watson1978: When I preview this in local environment, it shows as following
image

deployment/source-only-mode.md
daipom: Thanks, but behavior may be different from the production environment.

At least, the anchor would work because the following works now.

* https://docs.fluentd.org/output/file#less-than-buffer-greater-than-section
* https://github.com/fluent/fluentd-docs-gitbook/blob/2483910bbf16c723d052b38d92b7e43e2bfce2e7/output/file.md#L28

I cannot be sure about the broken description because there is no other similar section.
I will remove the `<` and `>` just in case.

deployment/source-only-mode.md
```diff
@@ -0,0 +1,76 @@
+# Source Only Mode
+
+Since v1.18.0, Fluentd can launch with source-only mode.
+(Not supported on Windows)
+
+* In this mode, only input plugins runs.
```
cosmo0920:
```suggestion
* In this mode, only input plugins run.
```

This is because input plugins are not 3rd person singular.

deployment/source-only-mode.md
daipom: Do you mean this is confusing with right/left?
Certainly, it is.
I will fix this.

deployment/source-only-mode.md
daipom: Thanks, I will fix `recovering` to `recovery` entirely. => Fixed: de81205d8661b7d8e1a3ac738c39863536a68968

What is the problem with the following expression?

* `with the temporary buffer left`

deployment/source-only-mode.md
```diff
@@ -0,0 +1,76 @@
...
+## How to launch Fluentd with source-only mode
+
+You can launch Fluentd with source-only mode in the following ways.
+
+* [Command Line Option - --with-source-only](command-line-option.md)
+* [System Configuration - with_source_only](system-config.md#with_source_only)
+
+## Temporary file buffer
+
+During source-only mode, the events are stored in a temporary file buffer.
```
cosmo0920: What does the events actually mean?

The events should be ingested and stored. So, I'd like to recommend to describe more concretely like:

```suggestion
During source-only mode, the ingested events are stored in a temporary file buffer.
```

deployment/system-config.md
daipom: https://github.com/fluent/fluentd-docs-gitbook/pull/527#discussion_r1914534584

deployment/source-only-mode.md
```diff
@@ -0,0 +1,76 @@
...
+the supervisor process to cancel with-source-only mode and process data.
+```
+
+This file buffer is the buffer of [out_buffer](../output/buffer.md).
+It works with the default settings of `out_buffer` except for the following points:
+
+* `path` is automatically determined by default.
+* `overflow_action` is `drop_oldest_chunk` by default.
+
+If need, you can configure some options for the buffer in System Configuration.
```
cosmo0920: Just following the sentence structure, we need to change as follows:

```suggestion
If needed, you can configure some options for the buffer in System Configuration.
```

deployment/source-only-mode.md
daipom: Fixed: de81205d8661b7d8e1a3ac738c39863536a68968

deployment/source-only-mode.md
daipom: Fixed: de81205d8661b7d8e1a3ac738c39863536a68968

deployment/source-only-mode.md
cosmo0920: Including @Watson1978 's suggestion, I'm also recommend to use as follows here:

```suggestion
If this recovery is necessary, i.e., Fluentd stops with the temporary buffer left, the following warning log will be displayed.
```

output/buffer.md
```diff
@@ -0,0 +1,114 @@
...
+    </buffer>
+  </match>
+</label>
+
+<match foo.**>
+  @type stdout
+</match>
+```
+
+In the above example, events emitted by `in_udp` are once stored in the buffer of this plugin, then re-routed and output by `out_stdout`.
```
cosmo0920: emit could be used for sending outside of Fluentd.
So, we need to use ingested instead of emited in this case.
```suggestion
In the above example, events ingested by `in_udp` are once stored in the buffer of this plugin, then re-routed and output by `out_stdout`.
```

cosmo0920 (commented):
I commented several concerns at the first glance.

deployment/source-only-mode.md
cosmo0920: Output is also able to use as verb in this case. But this type of usages is not common. So, I recommend to use display as a verb here. In this context, display should be changed its form as a past participle.

daipom:
@cosmo0920 
Thanks for your review! Fixed!

deployment/source-only-mode.md
```diff
@@ -0,0 +1,76 @@
...
+* [System Configuration - with_source_only](system-config.md#with_source_only)
+
+## Temporary file buffer
+
+During source-only mode, the ingested events are stored in a temporary file buffer.
+After `SIGWINCH` is sent to the supervisor and this mode is canceled, this buffer begins to load.
+
+By default, the file buffer path is as follows.
+
+* `/tmp/fluent/source-only-buffer/{Unique ID for the Fluentd instance}`
```
kenhys, line 25: It might be better to clarify "What instance".
e.g supervisor, worker,  plugin and so on.

deployment/source-only-mode.md
kenhys: or `remained`?

deployment/system-config.md
```diff
@@ -100,6 +100,20 @@ Logs event verbosely.
...
+| :--- | :--- | :--- |
+| bool | nil | 1.18.0 |
+
+Not supported on Windows.
+
+Invokes a fluentd only with input plugins.
+
+See [Source Only Mode](source-only-mode.md) for details.
+
+See also [source_only_buffer section](#less-than-source_only_buffer-greater-than-section).
```
kenhys, line 115: It it intended for `#sourceonlybuffer-section`

deployment/source-only-mode.md
daipom: It is a common ID for the supervisor and all workers belonging to the supervisor, but I can't explain it briefly...

deployment/system-config.md
daipom: https://github.com/fluent/fluentd-docs-gitbook/pull/527#discussion_r1914534584

> Thanks, but behavior may be different from the production environment.
> 
> At least, the anchor would work because the following works now.
> 
>  * https://docs.fluentd.org/output/file#less-than-buffer-greater-than-section
>  * https://github.com/fluent/fluentd-docs-gitbook/blob/2483910bbf16c723d052b38d92b7e43e2bfce2e7/output/file.md#L28

kenhys: Oh, I've overlooked....

deployment/source-only-mode.md
kenhys: Hmm, so let's keep it as is.

deployment/source-only-mode.md
daipom: > What is the problem with the following expression?
> 
>     * `with the temporary buffer left`

Fixed! I use `remained` instead of `left`.

deployment/source-only-mode.md
daipom: Fixed! I use `remained`!

kenhys (approved):
LGTM.

daipom:
Thanks for your review!!

PR #528, opened 2025-01-15
zero-downtime restart: add initial document

daipom:
https://github.com/fluent/fluentd/pull/4624


deployment/rpc.md
```diff
@@ -29,10 +29,16 @@ As evident from the output above, each endpoint returns a JSON object as its res
...
+| `/api/processes.zeroDowntimeRestart` | [SIGUSR2](signals.md#sigusr2) | Restarts Fluentd with zero-downtime. |
 | `/api/processes.flushBuffersAndKillWorkers` | [SIGUSR1](signals.md#sigusr1) and [SIGTERM](signals.md#sigint-or-sigterm) | Flushes buffer and stops the daemon. |
 | `/api/plugins.flushBuffers` | [SIGUSR1](signals.md#sigusr1) | Flushes the buffered messages. |
-| `/api/config.gracefulReload` | [SIGUSR2](signals.md#sigusr2) | Reloads configuration. |
 | `/api/config.reload` | [SIGHUP](signals.md#sighup) | Reloads configuration. |
+| `/api/config.gracefulReload` | --- | Reloads configuration. |
+
+Appendix:
+
+* `/api/processes.zeroDowntimeRestart`: This is supported since v1.18.0 on non-Windows.
```
kenhys: How about adding version column to above table?
or mention about available version in explanation because "which version?" is important information.

deployment/signals.md
```diff
@@ -18,15 +18,56 @@ Forces the buffered messages to be flushed and reopens Fluentd's log. Fluentd wi
...
+
+**Comparison with SIGHUP**
+
+`SIGHUP` gracefully restarting the worker process to reload.
+
+This method does not cause socket downtime, so if there is no need to restart the supervisor, `SIGHUP` is a lighter zero-downtime restart method.
+
+**Comparison with Graceful reload**
+
+You can still use Graceful reload feature by sending `SIGUSR2` directly to the worker processor or using [RPC](rpc.md) even after v1.18.0.
```
kenhys:
```suggestion
You can still use Graceful reload feature by sending `SIGUSR2` directly to the worker process or using [RPC](rpc.md) even after v1.18.0.
```

deployment/zero-downtime-restart.md
```diff
@@ -0,0 +1,48 @@
...
+4. Launch new workers, and stop old processes in parallel.
+   * Launch new workers with [Source Only Mode](source-only-mode.md).
+     * In addition to source-only mode, further limit the starting pluings to only those that support this feature.
+     * Data received by the new workers are stored in the temporary buffer of source-only mode.
+   * Send `SIGTERM` to the old supervisor after `10s` delay.
+5. The old supervisor stops and sends `SIGWINCH` to the new one.
+6. The new workers starts to run fully.
+   * The temporary buffer of source-only mode starts to load.
+
+You can configure the temporary buffer.
```
kenhys: The context "when need to be configured" seems missing. 
A bit more explanation is needed here.

plugin-development/api-plugin-input.md
```diff
@@ -102,7 +102,31 @@ router.emit(tag, time, {:foo => 'bar'})
...
+```
+
+To do this, the following condition must be met:
+
+* This plugin can run in parallel with another Fluentd.
+
+This is because there is a period when the old process and the new process run in parallel during a zero-downtime restart.
+
+After addressing the following considerations and ensuring there are no issues, override this method.
+Then, the plugin will not experience downtime with zero-downtime restart.
```
kenhys:
```suggestion
Then, the plugin will succeed with zero-downtime restart.
```

It could be simple enough?

kenhys (changes requested):

deployment/rpc.md
daipom: Fixed!
03e1b038c876b28da1cb7e712aa41d0e02c40925

deployment/zero-downtime-restart.md
daipom: Is it something like this?

```diff
- You can configure the temporary buffer.
+ You can configure the temporary buffer in advance.
```

plugin-development/api-plugin-input.md
daipom: Thanks!
I have a little concern about whether the “succeed” is appropriate here, but I think it makes sense, so let's go with that!

deployment/zero-downtime-restart.md
daipom: I see! Certainly, it was hard to understand the context of the buffer.

I noticed it has been already explained in https://docs.fluentd.org/deployment/source-only-mode#temporary-file-buffer.
So, I fixed this as 17d31346d17c4164582e2c7f559002269d0a41a7.
How about this?

deployment/signals.md
```diff
@@ -18,15 +18,57 @@ Forces the buffered messages to be flushed and reopens Fluentd's log. Fluentd wi
...
+
+#### Graceful reload
+
 Reloads the configuration file by gracefully re-constructing the data pipeline. Fluentd will try to flush the entire memory buffer at once, but will not retry if the flush fails. Fluentd will not flush the file buffer; the logs are persisted on the disk by default.
 
-This signal has been supported since v1.9.0.
+Limitations:
+
+* A change to System Configuration (`<system>`) is ignored.
+* All plugins must not use class variable when restarting.
```
Watson1978:
```suggestion
* All plugins must not use class variable.
```

daipom: Thanks!

Watson1978 (approved):

kenhys (approved):
LGTM.

PR #529, opened 2025-01-17
Restructure quickstart section

kenhys:
There is no need to know about older versions of Fluentd or td-agent, so move it from quickstart section.

It is still useful for migration, so keep it as appendix section.


kenhys:
image

image

image

daipom (approved):
Thanks!
LGTM! We should restructure these old info.

PR #534, opened 2025-01-20
Emphasize EOL information

kenhys:
Change previous 'info' level to 'danger' to raise an attention.


daipom (approved):
Thanks!

kenhys:
NOTE: honkit theme  it bit different color.

Before:

image

After:

image

PR #537, opened 2025-01-21
Fix wording in hints about deprecated td-agent

kenhys:
* some content in info should be now danger.
* note missing fluent-package on macOS


daipom (approved):
LGTM. Thanks!

PR #542, opened 2025-01-21
installation: refactor filepaths

daipom:
* The actual link and file path will match, so this will reduce confusion.
* This allows us to add content about `fluent-package` in https://docs.fluentd.org/installation/install-fluent-package in the future.

This does not affect the views.


kenhys (commented):
It is good to refactor under sub-directory, but directory name seems a bit redundant.

How about changing like this?

installation/calyptia-fluentd/...
installation/fluent-package/...

daipom:
The problem is that it will change the current URL.
Currently, the URL is as follows.

* https://docs.fluentd.org/installation/install-fluent-package/install-by-rpm-fluent-package

If we prefer not to change this, we need to make the name `install-fluent-package`.

On the other hand, since this URL was created recently, there might be a way of thinking like, "Let's change it."

What do you think?

kenhys (approved):
LGTM.

installation/install-fluent-package/... URL will be assigned automatically, 
it is better not to shorten. let it gitbook.

daipom:
Thanks!

PR #546, opened 2025-01-22
Update hdfs howto for fluent-package

kenhys:
NOTE: It might better about recent CDH information later.

See https://community.cloudera.com/t5/Support-Questions/is-there-any-open-source-CDH-version-available-to-download/td-p/316250


daipom:
Thanks!
Should we update here too?

https://github.com/fluent/fluentd-docs-gitbook/blob/2897259b7bae403e353f337f9c96fb53a13eec94/how-to-guides/http-to-hdfs.md?plain=1#L119

kenhys:
Thanks!

daipom (approved):
Thanks!

PR #555, opened 2025-01-23
quickstart/fluent-package: fix expression

daipom:
https://docs.fluentd.org/quickstart/fluent-package-v5-vs-td-agent#conclusion-use-fluent-package-v5-instead-of-td-agent-v4


daipom:
* Fixed in #556

PR #560, opened 2025-01-31
parser_json: use JSON as fallback parser instead of Yajl

Watson1978:
Related to https://github.com/fluent/fluentd/pull/4813


daipom:
Thanks!

How about adding the mention of the things before 1.19.0?

Watson1978:
Thanks.
I added the NOTE,
```
(The fallback has been changed from `yajl` to `json` since v1.19.0.)
```

daipom (approved):
Thanks!!

Issue #565, opened 2025-02-06
The internal links inside installation/install-fluent-package were broken


livekn:
### Describe the bug

The internal links were broken because it's inside "/installation/install-fluent-package", so should use "../../" instead of "../"

![Image](https://github.com/user-attachments/assets/3398dcf1-7b09-49a6-b589-ef60b4e42ccd)

### Link to the problematic documentation

https://docs.fluentd.org/installation/install-fluent-package/install-by-rpm-fluent-package
https://docs.fluentd.org/installation/install-fluent-package/install-by-deb-fluent-package
https://docs.fluentd.org/installation/install-fluent-package/install-by-dmg-fluent-package
https://docs.fluentd.org/installation/install-fluent-package/install-by-msi-fluent-package

### Expected explanation

The link should work

kenhys:
Thanks!
If you afford to fix it, Pull request is wellcome.

livekn:
Update: all the internal links inside these pages were broken. I fix them all

daipom:
Thanks so much!

PR #566, opened 2025-02-06
Fix broken links

livekn:
Fix the broken links inside /installation/install-fluent-package (#565)


livekn:
Update: All the internal links in these pages with that problem, I fix them all

daipom (approved):
It's an oversight from when I changed the directory structure.
Thank you for fixing it!

PR #567, opened 2025-02-11
Fix various typos

kenhys:


kenhys:
Strictly re-use pass-thru is not so bad, but unify to more common expression.

Watson1978 (approved):

daipom (approved):
Thanks!

PR #570, opened 2025-02-17
Fix broken links

shusch:
Fix relative path links in `installation/install-fluent-package/install-by-rpm-fluent-package.md`

https://docs.fluentd.org/installation/install-fluent-package/install-by-rpm-fluent-package


daipom (approved):
Thanks!!

PR #571, opened 2025-02-21
Document double asterisk globbing for path on tail input

okin:
My search for this did not bring up good results, so I wanted to add documentation to save others the hassle.


daipom (approved):
Thanks!

PR #575, opened 2025-04-10
http_server: stop fallback to WEBrick

daipom:
https://github.com/fluent/fluentd/pull/4899

https://docs.fluentd.org/plugin-helper-overview/api-plugin-helper-http_server


Watson1978 (approved):
👍🏻

PR #580, opened 2025-05-26
configuration: update config file location

daipom:
https://docs.fluentd.org/configuration/config-file#config-file-location


kenhys (approved):
LGTM.

daipom:
Thanks for your review!

PR #584, opened 2025-06-10
Fix dead link

Watson1978:
Looks like some links move us to 404 page such as following.

スクリーンショット 2025-06-10 12 20 51


daipom (approved):
Thanks!

PR #585, opened 2025-06-10
in_tail: add description of bug in encoding option

daipom:
* Page: https://docs.fluentd.org/input/tail#encoding-from_encoding
* Bug: https://github.com/fluent/fluentd/issues/4863#issuecomment-2957719361


Watson1978 (approved):
👍🏻

PR #586, opened 2025-06-11
Fix link contains fragment

Watson1978:
When linking to a relative path that includes a fragment, 
it seems that the file extension should be included.


kenhys (approved):
LGTM.

PR #588, opened 2025-06-24
system-config: some options should not be under <log> section

daipom:
https://docs.fluentd.org/deployment/system-config#enable_input_metrics
https://docs.fluentd.org/deployment/system-config#enable_size_metrics

These options are not under `<log>` section.

https://github.com/fluent/fluentd/blob/6821e4bb4b80bf99473541dc6a5779dbc3784c70/lib/fluent/system_config.rb#L54-L55


kenhys:
Thanks!

Oh...
It was introduced by https://github.com/fluent/fluentd-docs-gitbook/pull/358

kenhys (approved):
LGTM.

PR #590, opened 2025-06-27
buffer-section: fix inaccurate description about compress option

daipom:
https://docs.fluentd.org/configuration/buffer-section#buffering-parameters

It is `Chunk#write_to` to decompress the chunk automatically.

The current description sounds like it has already been decompressed before `Output#write`.


kenhys (approved):
LGTM.

PR #592, opened 2025-06-27
api-config-types: add note about default value

Watson1978:
Related to https://github.com/fluent/fluent-plugin-opensearch/pull/159


daipom (approved):
Thanks!

PR #593, opened 2025-07-29
out_stdout: add use_logger option

daipom:
https://docs.fluentd.org/output/stdout

Corresponding fix:

* https://github.com/fluent/fluentd/pull/5036


Watson1978 (approved):
👍🏻

PR #597, opened 2025-08-15
out_file: add symlink_path_use_relative

Watson1978:
`symlink_path_use_relative` parameter was introduced at Fluentd 1.19

https://github.com/fluent/fluentd/pull/4904


daipom (approved):
Thanks!!

PR #602, opened 2025-12-04
output: add workaround explanation with stream_buffer_size

kenhys:
It might be better to add explanation about workaround.

See https://github.com/fluent/fluentd/issues/5133


Watson1978 (approved):
Thanks!

PR #603, opened 2025-12-13
out_file: add example explanation

sye-lens:
Because the output files generated by running "example configuration" are difficult to understand.


daipom (approved):
Thanks!

PR #606, opened 2025-12-24
Update EOL information about v5

kenhys:


kenhys:
Before: gitbook rendering

image

After: honkit rendering

image

NOTE: Ignore background color because there is incompatibility between  honkit and gitbook.

Watson1978 (approved):
👍🏻

PR #607, opened 2026-01-08
Fix config-file-yaml

Watson1978:


configuration/config-file-yaml.md
```diff
@@ -313,41 +313,41 @@ The element in separate configuration files can be imported using the **!include
 ```yaml
 config:
   # Include config files in the ./config.d directory
-  - !include config.d/*.conf
+  - !include config.d/*.yaml
```
Watson1978, line 316: Currently, `!include` directives include only YAML formatted configuration files.

Therefore, files with the .conf extension may cause confusion with regular configuration files.

We already use this `!include` pattern in
https://github.com/fluent/fluentd-docs-gitbook/blob/3f02b58dde45fff175e648093ce10b71ed22b508/configuration/config-file-yaml.md?plain=1#L365

configuration/config-file-yaml.md
```diff
@@ -49,7 +49,8 @@ Under `config` object, Fluentd will handle the following elements:
 6. **`$arg`** defines directive **arg**'s equivalent objects (e.g. \<DIRECTIVE **arg**\>)
 7. **`$name`** defines label directive **name** equivalent objects (e.g. \<label **name**\>)
 8. **`$type`** specifies `@type` or `type` for instantiating plugin type
-9. **`$log_level`** specifies `@log_level` per plugin
+9. **`$id`** specifies plugin id
```
Watson1978, line 52: There is implementations to handle $id.

Ref. https://github.com/fluent/fluentd/blob/931f22a76158e77f7f551f6f7943bb7346973af3/lib/fluent/config/yaml_parser/parser.rb#L137-L139

kenhys (approved):
LGTM.

PR #609, opened 2026-02-25
Navigate to buffer section

kenhys:
It might be better to navigate to buffer section because it explains some chunk key use-case.


Watson1978 (approved):
👍🏻

PR #626, opened 2026-07-14
Fix broken link to list of plugins

manu-cyber:
https://fluentd.org/plugin/ returns a (fake) 404 while https://www.fluentd.org/plugins/ is the working URL for the list of plugins


Watson1978 (approved):
Thanks

PR #630, opened 2026-07-17
output: apply to change `extract_placeholders()` API behavior

Watson1978:
https://github.com/fluent/fluentd/pull/5391

### DIFF
--- a/configuration/buffer-section.md
+++ b/configuration/buffer-section.md
@@ -280,7 +280,7 @@ Any key is acceptable as a chunk key. If a key other than specified in the chunk
 
 ### Path Boundary Validation
 
-Since v1.19.3, Fluentd validates the values substituted into `${tag}` and `${key}` placeholders so that they cannot cross an unintended path boundary. Because tags and record fields come from incoming events, an unexpected event could otherwise make a plugin like `out_file` create or access a file outside of the intended directory.
+Since v1.19.3, Fluentd validates the values substituted into `${tag}` and `${key}` placeholders to prevent unintended path traversal. Because tags and record fields come from incoming events, an unexpected event could otherwise make a plugin like `out_file` create or access a file outside of the intended directory.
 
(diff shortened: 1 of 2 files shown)


configuration/buffer-section.md
```diff
@@ -278,6 +278,29 @@ Any key is acceptable as a chunk key. If a key other than specified in the chunk
 </match>
 ```
 
+### Path Boundary Validation
+
+Since v1.19.3, Fluentd validates the values substituted into `${tag}` and `${key}` placeholders so that they cannot cross an unintended path boundary. Because tags and record fields come from incoming events, an unexpected event could otherwise make a plugin like `out_file` create or access a file outside of the intended directory.
```
kenhys:
```suggestion
Since v1.19.3, Fluentd validates the values substituted into `${tag}` and `${key}` placeholders to prevent unintended path traversal. Because tags and record fields come from incoming events, an unexpected event could otherwise make a plugin like `out_file` create or access a file outside of the intended directory.
```

How about using known terms here?

plugin-development/api-plugin-output.md
```diff
@@ -346,6 +346,10 @@ It extracts placeholders in the given string using the values in chunk.
 
 Return value is a `String`.
 
+Since v1.19.3, this method raises `Fluent::UnrecoverableError` if a value substituted into `${tag}` or `${key}` would cross a path boundary, e.g. a tag containing `../` or beginning with `/`. See [Path Boundary Validation](../configuration/buffer-section.md#path-boundary-validation).
```
kenhys:
```suggestion
Since v1.19.3, this method raises `Fluent::UnrecoverableError` if a value substituted into `${tag}` or `${key}` attempts a path traversal, e.g. a tag containing `../` or beginning with `/`. See [Path Boundary Validation](../configuration/buffer-section.md#path-boundary-validation).
```

kenhys (changes requested):

kenhys (approved):

PR #640, opened 2026-07-23
Add delimiter parameter of in_tcp

Watson1978:
The parameter is defined at lib/fluent/plugin/in_tcp.rb:45-46

    desc 'The payload is read up to this character.'
    config_param :delimiter, :string, default: "\n"

but input/tcp.md only mentioned it in a comment of the example configuration, with no section under "Parameters".

It is matched with String#index (in_tcp.rb:127), not as a pattern, so a multi-character delimiter works; this was verified by sending '{"k":"a"}||{"k":"b"}||' with `delimiter "||"`.

The interaction with message_length_limit follows in_tcp.rb:153-162, where the buffer is cleared and `discard_till_next_delimiter` is set.

The parameter was introduced at https://github.com/fluent/fluentd/pull/368

### DIFF
--- a/input/tcp.md
+++ b/input/tcp.md
@@ -141,6 +141,29 @@ You will get something like below:
 
 The maximum number of bytes for the message.
 
+### `delimiter`
+
+| type | default | version |
+| :--- | :--- | :--- |
+| string | `"\n"` \(newline\) | 0.10.51 |
+
+The payload is read up to this character. Since a TCP stream has no message boundary, `in_tcp` splits the received bytes at every occurrence of this string and passes each part to the `<parse>` section. The delimiter itself is not included in the parsed message.
+
+This is not a pattern but a plain string, so a delimiter of two or more characters can also be specified:
+
+```text
+<source>
+  @type tcp
+  tag tcp.events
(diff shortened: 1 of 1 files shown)


input/tcp.md
```diff
@@ -141,6 +141,29 @@ You will get something like below:
...
+
+The payload is read up to this character. Since a TCP stream has no message boundary, `in_tcp` splits the received bytes at every occurrence of this string and passes each part to the `<parse>` section. The delimiter itself is not included in the parsed message.
+
+This is not a pattern but a plain string, so a delimiter of two or more characters can also be specified:
+
+```text
+<source>
+  @type tcp
+  tag tcp.events
+  delimiter "||"
```
kenhys:
```suggestion
  delimiter "||"  # It will work expectedly if it splits with '||' against incoming "foo||bar", not work for 'foo|bar'.
```

kenhys: It might be better to describe expected use-case and exceptional one.

kenhys (changes requested):

kenhys (approved):

PR #653, opened 2026-07-28
docs: use https for fluentd.org links

Watson1978:
Replace `http://` with `https://` for all `www.fluentd.org` and `fluentd.org` URLs across the docs. This is mostly the CNCF footer boilerplate that appears on every page, plus a handful of in-body links. fluentd.org serves all of these over https, so the plain-http links were needless.


kenhys (approved):
good catch!

Issue #659, opened 2026-07-30
docs.fluentd.org (via GitBook) is out-of-sync


kenhys:
### Describe the bug

Recently (since July 21), docs.fluentd.org was not updated via webhooks.

fluentd-docs-gitbook 1.0 branch should be deployed via webhook, but not work for now.

<img width="838" height="208" alt="Image" src="https://github.com/user-attachments/assets/3bf64523-aa0c-4df7-8242-b7d0e97fc25c" />

It seems broken now.

### Link to the problematic documentation

All of 1.0 branch contents.

### Expected explanation

1.0 branch should be deployed via gitbook webhook

### Additional context

Target gitbook space IDs which are bound to might be either:

* `-MbEiFWqPnv1y_0Eq2Vw` (current?)
* `-LR7OsqPORtP86IQxs6E` (deprecated?)

kenhys:
@edsiper @agup006 Could you take a look at, please?

kenhys:
I've discussed with @Watson1978, concluded that it might be better to manage under CNCF umbrella. (as same as fluentd-website do)

kenhys:
I've contacted via CNCF servicedesk.

kenhys:
Now migrated from gitbook.com to Netlify.

PR #663, opened 2026-08-04
0.12: fix Template render error: with honkit

kenhys:
{{.foo}} syntax causes unexpected token:. error.


kenhys:
This could be blocker of #659

Watson1978 (approved):

PR #665, opened 2026-08-05
Add parameters which were missing from the docs

Watson1978:
Six parameters are accepted by the plugins but were not described anywhere in the documentation. Users could only find them by reading the plugin source.

* formatter/json.md   json_parser      (0.14.0)
* input/exec.md       encoding         (1.17.1)
* input/forward.md    backlog          (0.10.36)
* input/syslog.md     source_host_key  (0.10.53)
* input/tail.md       path_delimiter   (1.9.1)
* parser/multiline.md unmatched_lines  (1.9.3)

The version column is taken from the commit which introduced each `config_param` and the first release tag containing it, not guessed.

Two related fixes on pages touched here:

* input/forward.md: `resolve_hostname` cannot be `false` when `source_hostname_key` is set. The wording matches the note which input/syslog.md already has for the same constraint.
* input/syslog.md: `include_source_host` assigns `@source_address_key`, so it adds the source *address* under the field name given by `source_host_key`, not the hostname.

### DIFF
--- a/formatter/json.md
+++ b/formatter/json.md
@@ -9,6 +9,18 @@ By default, `json` formatter result doesn't contain `tag` and `time` fields.
 * [Common Parameters](../configuration/plugin-common-parameters.md)
 * [Format section configurations](../configuration/format-section.md)
 
+### `json_parser`
+
+| type | default | version |
+| :--- | :--- | :--- |
+| string | oj | 0.14.0 |
+
+Sets the library used to generate JSON. Despite its name, this parameter is for generating JSON, not for parsing it.
+
+If `oj` is specified, the [`oj`](https://github.com/ohler55/oj) gem is used. The gem must be installed separately because Fluentd does not require it by default. If it is not installed, the `json` standard library is used as a fallback and it is reported in the log at the info level.
+
+Any other value, including `yajl`, selects the `json` standard library. Unlike [`json_parser`](../parser/json.md#json_parser) of the `json` parser plugin, this parameter is not restricted to a fixed list of values.
+
 ### `add_newline`
 
 | type | default | version |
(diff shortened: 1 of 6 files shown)


formatter/json.md
```diff
@@ -9,6 +9,18 @@ By default, `json` formatter result doesn't contain `tag` and `time` fields.
 * [Common Parameters](../configuration/plugin-common-parameters.md)
 * [Format section configurations](../configuration/format-section.md)
 
+### `json_parser`
+
+| type | default | version |
+| :--- | :--- | :--- |
+| string | oj | 0.14.0 |
```
kenhys:
```suggestion
| string | oj | 0.12.19 |
```

gg json_parser fluentd-0.12.1{8,9} G config_param
fluentd-0.12.19/lib/fluent/parser.rb:      config_param :json_parser, :string, :default => 'oj'
fluentd-0.12.19/lib/fluent/formatter.rb:      config_param :json_parser, :string, :default => 'oj'

input/syslog.md
```diff
@@ -294,7 +294,17 @@ The field name of the client's address. If set, the client's address will be set
...
+
+### `source_host_key`
+
+| type | default | version |
+| :--- | :--- | :--- |
+| string | `source_host` | 0.10.53 |
+
+The field name used when `include_source_host` is `true`. This parameter has no effect on its own.
+
+Since `include_source_host` is deprecated, use `source_address_key` instead. Specifying `source_address_key` and `include_source_host` at the same time causes a configuration error.
```
kenhys:
```suggestion
Since 0.14.11, `include_source_host` is deprecated, use `source_address_key` instead. Specifying `source_address_key` and `include_source_host` at the same time causes a configuration error.
```

kenhys (changes requested):

kenhys (approved):

PR #669, opened 2026-08-07
Drop unused addcssjs plugin and googleanalytics.js

kenhys:
The version of UA- google analytics service
had already terminated and it is not used
because identifier is not customized.


Watson1978 (approved):
👍🏻

PR #670, opened 2026-08-07
Drop Ruby 3.2 support

Watson1978:
https://github.com/fluent/fluentd/pull/5465


kenhys (approved):
LGTM.

but pending until v1.20 release

PR #671, opened 2026-08-08
Honkit migration

### LINKED ISSUE #659: docs.fluentd.org (via GitBook) is out-of-sync

kenhys:
To fix #659, migrate to Netlify as same as fluentd-website.

* Use honkit to reduce migration cost
* Best effort  to keep compatibility with gitbook layout
* Use _redirect to replace .gitbook.yaml redirects.
* Simplify footer with localized-footer plugin
* Introduce package-lock.json to hold them
* Specify [[redirects]] to show image resources in netlify.toml


Watson1978 (approved):
These can be fixed later, so I don't think they should block merging this PR.

### 0.12

- The images are not displayed.
  - They are embedded as ``, so the generated pages request `/0.12/.gitbook/assets/...`. The redirect rule in `netlify.toml` only matches the site root (`from = "/.gitbook/assets/*"`), so nothing rewrites the versioned paths and they end up as 404. The files themselves are in place: `/0.12/assets/apache-to-mongodb.png` returns 200.
  - Ref. https://deploy-preview-671--fluentd-docs.netlify.app/0.12/articles/apache-to-mongodb.html
  - Adding rules for the sub-books would fix it:

    ```toml
    [[redirects]]
      from = "/0.12/.gitbook/assets/*"
      to = "/0.12/assets/:splat"
      status = 200

    [[redirects]]
      from = "/1.0/.gitbook/assets/*"
      to = "/1.0/assets/:splat"
      status = 200
    ```

- The 0.12 pages are titled `<title>Getting Started · Fluentd 1.0 Documentation</title>`. It would be better to show `0.12` here. This seems to come from `scripts/build.sh`, which copies the 1.0 `book.json` into the 0.12 worktree and overwrites its own `"title": "Fluentd v0.12 Documentation"`.
- The raw markdown files are deployed, which is unnecessary.
  - Ref. https://deploy-preview-671--fluentd-docs.netlify.app/_build/0.12/input/tail.md (the same applies to `/_build/1.0/`)
  - The `_build/` worktrees are created before `npx honkit build` runs, and honkit does not ignore them, so both branches are copied into the output as assets. Adding `_build` to `.bookignore` should be enough.

kenhys:
Thanks, I'll fix them in another PR.

PR #673, opened 2026-08-10
Fix in-page anchors broken by the HonKit migration

Watson1978:
HonKit builds heading ids differently from GitBook: it drops backticks, `<`, `>` and `_` from the heading text. The links written for GitBook therefore stopped resolving once #671 and #672 made HonKit the renderer for the published site.

| heading                     | GitBook id                             | HonKit id      |
| --------------------------- | -------------------------------------- | -------------- |
| `### `add_http_headers``    | add_http_headers                       | addhttpheaders |
| `### `<inject>` Section`    | less-than-inject-greater-than-section  | inject-section |

15 links across 6 files, 8 distinct anchors:

* input/http.md               #add_http_headers, #decompression_size_limit
* input/syslog.md             #source_host_key
* output/http.md              #allowed_hosts
* output/exec_filter.md       #less-than-{inject,extract}-greater-than-section
* output/file.md              #less-than-buffer-greater-than-section
* deployment/system-config.md #less-than-source_only_buffer-greater-than-section

Verified by building the site with the repository's own configuration and matching every `href="#..."` against the `id=` and `name=` attributes on the same page. Before this change 8 of the 2366 in-page anchors did not resolve; after it, none are left.

The new ids were read out of the generated HTML rather than derived from the rule by hand.

### DIFF
--- a/deployment/system-config.md
+++ b/deployment/system-config.md
@@ -112,7 +112,7 @@ Invokes a fluentd only with input plugins.
 
 See [Source Only Mode](source-only-mode.md) for details.
 
-See also [source_only_buffer section](#less-than-source_only_buffer-greater-than-section).
+See also [source_only_buffer section](#sourceonlybuffer-section).
 
 ### `rpc_endpoint`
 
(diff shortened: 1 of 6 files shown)


kenhys (approved):
Good catch

PR #675, opened 2026-08-10
Replace default search plugins with search-plus

Watson1978:
The default search stack (search + lunr plugins) had two practical
problems on docs.fluentd.org:

* The generated search_index.json is about 6.7 MB (837 KB gzipped)
  and is fetched and parsed eagerly on every page view. Queries typed
  before loading completes throw "Search has not been initialized"
  with no visible error, and the typed query is never re-executed
  once the index becomes ready. As a result the search box appears
  broken right after page load. Since the length of this window
  depends on network and cache timing, the behavior also looks
  inconsistent across browsers.

* lunr applies an English stopword filter, so common words such as
  "for" always return "No results matching" even though they clearly
  appear in the documentation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "gitbook-plugin-partial-summary": "1.0.0",
     "gitbook-plugin-versions-select": "0.1.1",
     "honkit": "6.2.2",
-    "honkit-plugin-breadcrumbs": "0.0.3"
+    "honkit-plugin-breadcrumbs": "0.0.3",
+    "honkit-plugin-search-plus": "1.0.0"
   }
 }
--- a/book.json
+++ b/book.json
@@ -9,7 +9,8 @@
 		"intopic-toc",
 		"partial-summary",
 		"localized-footer",
-		"versions-select"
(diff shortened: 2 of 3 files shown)


kenhys:
What about honkit-plugin-search-plus?
https://github.com/dogatana/honkit-plugin-search-plus

Watson1978:
Replaced to [honkit-plugin-search-plus](https://github.com/dogatana/honkit-plugin-search-plus)

package.json
```diff
@@ -14,6 +14,7 @@
     "gitbook-plugin-intopic-toc": "1.1.1",
     "gitbook-plugin-localized-footer": "0.3.1",
     "gitbook-plugin-partial-summary": "1.0.0",
+    "gitbook-plugin-search-plus": "1.0.3",
```
kenhys: It should be honkit-plugin-search-plus 

https://www.npmjs.com/package/honkit-plugin-search-plus

Watson1978: Thanks

kenhys (approved):

PR #693, opened 2026-09-08
Point the v0.12 links at docs.fluentd.org

Watson1978:
The two links to the v0.12 manual still use the old GitBook domain. They redirect, but the rest of the docs now write v0.12 URLs as docs.fluentd.org/0.12, so use that form here as well.

The `in_forward` anchor fixed on the way as well.


kenhys (approved):
Good catch!

